PR #113, opened 2023-09-04
Fix incorrect 404 on pretranslations endpoints

### LINKED ISSUE #96: Serval incorrectly returns a 404 when there are no pretranslations
When the pretranslation endpoints are called and there are no pretranslations, then Serval incorrectly returns a 404 error. The `pretranslations` resource is a collection and should always exist as long as the engine exists. It is simply an empty collection, so it should return an empty array.

ddaspit:
- fixes #96

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/113)

### DIFF
--- a/src/Serval.Client/Client.g.cs
+++ b/src/Serval.Client/Client.g.cs
@@ -970,26 +970,15 @@ namespace Serval.Client
         /// <br/>  * The references defined in the SourceFile per line, if any.
         /// <br/>  * An auto-generated reference of `[TextId]:[lineNumber]`, 1 indexed.
         /// <br/>* **Translation**: the text of the pretranslation
+        /// <br/>            
+        /// <br/>Pretranslations can be filtered by text id if provided.
         /// </remarks>
         /// <param name="id">The translation engine id</param>
         /// <param name="corpusId">The corpus id</param>
+        /// <param name="textId">The text id (optional)</param>
         /// <returns>The pretranslations</returns>
         /// <exception cref="ServalApiException">A server side error occurred.</exception>
-        System.Threading.Tasks.Task<System.Collections.Generic.IList<Pretranslation>> GetAllPretranslationsAsync(string id, string corpusId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
-
-        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
-        /// <summary>
-        /// Gets all pretranslations from a TextId
-        /// </summary>
-        /// <remarks>
-        /// Similar to "get all pretranslations of a corpus," except that the results are filtered by TextId.
-        /// </remarks>
-        /// <param name="id">The translation engine id</param>
-        /// <param name="corpusId">The corpus id</param>
-        /// <param name="textId">The text id</param>
-        /// <returns>The pretranslations</returns>
-        /// <exception cref="ServalApiException">A server side error occurred.</exception>
-        System.Threading.Tasks.Task<System.Collections.Generic.IList<Pretranslation>> GetAllPretranslations2Async(string id, string corpusId, string textId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
+        System.Threading.Tasks.Task<System.Collections.Generic.IList<Pretranslation>> GetAllPretranslationsAsync(string id, string corpusId, string? textId = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
 
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         /// <summary>
@@ -2588,12 +2577,15 @@ namespace Serval.Client
         /// <br/>  * The references defined in the SourceFile per line, if any.
         /// <br/>  * An auto-generated reference of `[TextId]:[lineNumber]`, 1 indexed.
         /// <br/>* **Translation**: the text of the pretranslation
+        /// <br/>            
+        /// <br/>Pretranslations can be filtered by text id if provided.
         /// </remarks>
         /// <param name="id">The translation engine id</param>
         /// <param name="corpusId">The corpus id</param>
+        /// <param name="textId">The text id (optional)</param>
         /// <returns>The pretranslations</returns>
         /// <exception cref="ServalApiException">A server side error occurred.</exception>
-        public virtual async System.Threading.Tasks.Task<System.Collections.Generic.IList<Pretranslation>> GetAllPretranslationsAsync(string id, string corpusId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken))
+        public virtual async System.Threading.Tasks.Task<System.Collections.Generic.IList<Pretranslation>> GetAllPretranslationsAsync(string id, string corpusId, string? textId = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken))
         {
             if (id == null)
                 throw new System.ArgumentNullException("id");
@@ -2602,9 +2594,14 @@ namespace Serval.Client
                 throw new System.ArgumentNullException("corpusId");
 
             var urlBuilder_ = new System.Text.StringBuilder();
-            urlBuilder_.Append(BaseUrl != null ? BaseUrl.TrimEnd('/') : "").Append("/translation/engines/{id}/corpora/{corpusId}/pretranslations");
+            urlBuilder_.Append(BaseUrl != null ? BaseUrl.TrimEnd('/') : "").Append("/translation/engines/{id}/corpora/{corpusId}/pretranslations?");
             urlBuilder_.Replace("{id}", System.Uri.EscapeDataString(ConvertToString(id, System.Globalization.CultureInfo.InvariantCulture)));
             urlBuilder_.Replace("{corpusId}", System.Uri.EscapeDataString(ConvertToString(corpusId, System.Globalization.CultureInfo.InvariantCulture)));
+            if (textId != null)
+            {
+                urlBuilder_.Append(System.Uri.EscapeDataString("textId") + "=").Append(System.Uri.EscapeDataString(ConvertToString(textId, System.Globalization.CultureInfo.InvariantCulture))).Append("&");
+            }
+            urlBuilder_.Length--;
 
             var client_ = _httpClient;
             var disposeClient_ = false;
@@ -2662,136 +2659,6 @@ namespace Serval.Client
                         {
                             string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
                             throw new ServalApiException("The engine or corpus does not exist", status_, responseText_, headers_, null);
-                        }
-                        else
-                        if (status_ == 405)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The method is not supported", status_, responseText_, headers_, null);
-                        }
-                        else
-                        if (status_ == 409)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The engine needs to be built first", status_, responseText_, headers_, null);
-                        }
-                        else
-                        if (status_ == 503)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("A necessary service is currently unavailable. Check `/health` for more details. ", status_, responseText_, headers_, null);
-                        }
-                        else
-                        {
-                            var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The HTTP status code of the response was not expected (" + status_ + ").", status_, responseData_, headers_, null);
-                        }
-                    }
-                    finally
-                    {
-                        if (disposeResponse_)
-                            response_.Dispose();
-                    }
-                }
-            }
-            finally
-            {
-                if (disposeClient_)
-                    client_.Dispose();
-            }
-        }
-
-        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
-        /// <summary>
-        /// Gets all pretranslations from a TextId
-        /// </summary>
-        /// <remarks>
-        /// Similar to "get all pretranslations of a corpus," except that the results are filtered by TextId.
-        /// </remarks>
-        /// <param name="id">The translation engine id</param>
-        /// <param name="corpusId">The corpus id</param>
-        /// <param name="textId">The text id</param>
-        /// <returns>The pretranslations</returns>
-        /// <exception cref="ServalApiException">A server side error occurred.</exception>
-        public virtual async System.Threading.Tasks.Task<System.Collections.Generic.IList<Pretranslation>> GetAllPretranslations2Async(string id, string corpusId, string textId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken))
-        {
-            if (id == null)
-                throw new System.ArgumentNullException("id");
-
-            if (corpusId == null)
-                throw new System.ArgumentNullException("corpusId");
-
-            if (textId == null)
-                throw new System.ArgumentNullException("textId");
-
-            var urlBuilder_ = new System.Text.StringBuilder();
-            urlBuilder_.Append(BaseUrl != null ? BaseUrl.TrimEnd('/') : "").Append("/translation/engines/{id}/corpora/{corpusId}/pretranslations/{textId}");
-            urlBuilder_.Replace("{id}", System.Uri.EscapeDataString(ConvertToString(id, System.Globalization.CultureInfo.InvariantCulture)));
-            urlBuilder_.Replace("{corpusId}", System.Uri.EscapeDataString(ConvertToString(corpusId, System.Globalization.CultureInfo.InvariantCulture)));
-            urlBuilder_.Replace("{textId}", System.Uri.EscapeDataString(ConvertToString(textId, System.Globalization.CultureInfo.InvariantCulture)));
-
-            var client_ = _httpClient;
-            var disposeClient_ = false;
-            try
-            {
-                using (var request_ = new System.Net.Http.HttpRequestMessage())
-                {
-                    request_.Method = new System.Net.Http.HttpMethod("GET");
-                    request_.Headers.Accept.Add(System.Net.Http.Headers.MediaTypeWithQualityHeaderValue.Parse("application/json"));
-
-                    PrepareRequest(client_, request_, urlBuilder_);
-
-                    var url_ = urlBuilder_.ToString();
-                    request_.RequestUri = new System.Uri(url_, System.UriKind.RelativeOrAbsolute);
-
-                    PrepareRequest(client_, request_, url_);
-
-                    var response_ = await client_.SendAsync(request_, System.Net.Http.HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
-                    var disposeResponse_ = true;
-                    try
-                    {
-                        var headers_ = System.Linq.Enumerable.ToDictionary(response_.Headers, h_ => h_.Key, h_ => h_.Value);
-                        if (response_.Content != null && response_.Content.Headers != null)
-                        {
-                            foreach (var item_ in response_.Content.Headers)
-                                headers_[item_.Key] = item_.Value;
-                        }
-
-                        ProcessResponse(client_, response_);
-
-                        var status_ = (int)response_.StatusCode;
-                        if (status_ == 200)
-                        {
-                            var objectResponse_ = await ReadObjectResponseAsync<System.Collections.Generic.IList<Pretranslation>>(response_, headers_, cancellationToken).ConfigureAwait(false);
-                            if (objectResponse_.Object == null)
-                            {
-                                throw new ServalApiException("Response was null which was not expected.", status_, objectResponse_.Text, headers_, null);
-                            }
-                            return objectResponse_.Object;
-                        }
-                        else
-                        if (status_ == 401)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The client is not authenticated", status_, responseText_, headers_, null);
-                        }
-                        else
-                        if (status_ == 403)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The authenticated client cannot perform the operation or does not own the translation engine", status_, responseText_, headers_, null);
-                        }
-                        else
-                        if (status_ == 404)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The engine or corpus or text does not exist", status_, responseText_, headers_, null);
-                        }
-                        else
-                        if (status_ == 405)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The method is not supported", status_, responseText_, headers_, null);
                         }
                         else
                         if (status_ == 409)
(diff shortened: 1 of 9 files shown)


johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-Nd_kl9m5V3nTFHL8lRk:b-2wzd2y) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*
> ```Smalltalk
>     /// <response code="401">The client is not authenticated</response>
>     /// <response code="403">The authenticated client cannot perform the operation or does not own the translation engine</response>
>     /// <response code="404">The engine or corpus does not exist</response>
> ```

I don't understand - if they give a textId in the url and it does not exist, why wouldn't we use a 404 instead of a 409?  The data does not exist - which would be a 404...
"The HTTP 409 status code (Conflict) indicates that the request could not be processed because of conflict in the request, such as the requested resource is not in the expected state, or the result of processing the request would create a conflict within the resource."

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/113)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135)

___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 641 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_n4w10Qz_DTpsRzG8:-Nda1_3a0QszoFnwiiCw:b-cm1udg) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L641)):*

Why is the built check at the end rather than the beginning?
</blockquote></details>

I wasn't messing with the 409 checks, because they all need to be removed anyway. The engines should be performing these checks. Maybe I should just go ahead and remove them now.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-Nda1zJIBTrPUDA03zYg:bks80y0) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

I don't understand - if they give a textId in the url and it does not exist, why wouldn't we use a 404 instead of a 409?  The data does not exist - which would be a 404...
"The HTTP 409 status code (Conflict) indicates that the request could not be processed because of conflict in the request, such as the requested resource is not in the expected state, or the result of processing the request would create a conflict within the resource."
</blockquote></details>

I struggled with that question a bit when I was making this change. Unfortunately, there is not a good way to check if a text ID exists in a corpus. There are some data file formats that can hold multiple texts in a single file. For example, if the format is a Paratext backup, we will need to unzip the file to check if it contains the text.

For this endpoint, the text ID parameter is really a way to filter the pretranslations list. We could make this more explicit by making the text ID parameter part of the query string.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 641 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_n4w10Qz_DTpsRzG8:-NdaE_buA2E0OT4Ll7xb:b-wlvzog) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L641)):*

I wasn't messing with the 409 checks, because they all need to be removed anyway. The engines should be performing these checks. Maybe I should just go ahead and remove them now.
</blockquote></details>

How would the engine be performing the check if the API endpoint just does a mongo query?

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-NdaEzevE8u7ZEzhB8bf:b-6b25c2) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

I struggled with that question a bit when I was making this change. Unfortunately, there is not a good way to check if a text ID exists in a corpus. There are some data file formats that can hold multiple texts in a single file. For example, if the format is a Paratext backup, we will need to unzip the file to check if it contains the text.

For this endpoint, the text ID parameter is really a way to filter the pretranslations list. We could make this more explicit by making the text ID parameter part of the query string.
</blockquote></details>

The query string would make more sense - it could also be used to filter on books or chapters even if there is only one file.  That also appears that it would scale to multiple files.  Instead of removing or changing this endpoint, how about we extend the main translation suggestions endpoint to include an optional query?

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/113)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135)

___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 641 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_n4w10Qz_DTpsRzG8:-NdaHe5g8bkX5jJGi_sn:b-pp9i6c) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L641)):*

How would the engine be performing the check if the API endpoint just does a mongo query?
</blockquote></details>

There are cases where it would be perfectly valid for an engine to be able to respond to some of these endpoints even if it hasn't been built. Serval should defer to the engine to determine if a 409 should be returned or not. I will remove the checks and add an exception filter that returns 409 if a gRPC engine server responds with an Aborted (the equivalent of 409 for gRPC) status code. We will need to update the Machine engine server to perform the checks.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-NdaJkLW3D8e6hia6Gax:b-vqlbi0) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

The query string would make more sense - it could also be used to filter on books or chapters even if there is only one file.  That also appears that it would scale to multiple files.  Instead of removing or changing this endpoint, how about we extend the main translation suggestions endpoint to include an optional query?
</blockquote></details>

I agree. I will go ahead and make the change.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-NdaMV3O8B66RlmretT5:b-hstxsi) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

I agree. I will go ahead and make the change.
</blockquote></details>

Great - also I believe then that this function should check for the textId, as it will be a more scoped function (not be applicable for files in a paratext project, etc.).

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/113)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @johnml1135)

___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 641 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_n4w10Qz_DTpsRzG8:-NdbbV5p1ubgoKbHwtLD:b-yux9f1) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L641)):*

This logic is a few places - get wordgraph, get translation, get pretranslation.  Each are a little different.  The functions where we are querying Machine (get wordgraph, etc.), we can just rely on the function itself for the exception.  For pretranslations, should we add a new endpoint asking if pretranslations are available for this engine yet?
</blockquote></details>

We could add a property to the engine model that indicates if pretranslations are available. We would need to update the gRPC API to allow engines to communicate this. I don't think it is really worth the effort right now.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-NdbbUGCDWVosIUN1vkj:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

Great - also I believe then that this function should check for the textId, as it will be a more scoped function (not be applicable for files in a paratext project, etc.).
</blockquote></details>

Done.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 614 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeFqlT0WVQsB2wfhC8:-NdeFqlT0WVQsB2wfhC9:b-5n1t1s) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L614)):*
> ```Smalltalk
>         [NotNull] string id,
>         [NotNull] string corpusId,
>         [FromQuery] string? textId,
> ```

Should this be defaulted to null?  I don't know if it is automatic or not - but it should be explicit...

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 597 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeG1TR82VLD4q0gPFc:-NdeG1TR82VLD4q0gPFd:baa8omm) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L597)):*
> ```Smalltalk
>     /// <param name="id">The translation engine id</param>
>     /// <param name="corpusId">The corpus id</param>
>     /// <param name="textId">The text id</param>
> ```

The text (optional).

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 593 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeG6vZ9ZBrQWyEgLUX:-NdeG6vZ9ZBrQWyEgLUY:b-swankt) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L593)):*
> ```Smalltalk
>     /// * **Translation**: the text of the pretranslation
>     ///
>     /// Pretranslations can be filtered by text id.
> ```

...text id if provided.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-NdeGJ2A8pV2dVsH2bj2:bdsdccb) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

Done.
</blockquote></details>

Combining the 2 functions into 1 - I like it.  Is it truly backwards compatible though?  Wouldn't the syntax of the url be different?  I believe that Peter is already using the "get by textId" function...

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 612 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-NdeGeITCj2nUajpiAcT:-NdeGeITCj2nUajpiAcU:b-lcqfzj) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L612)):*
> ```Smalltalk
>     /// <response code="404">The engine or corpus does not exist</response>
>     /// <response code="405">The method is not supported</response>
>     /// <response code="409">The engine needs to be built first</response>
> ```

If we are going to remove this comment here - we should remove it everywhere.

johnml1135:
*[`src/Serval.Shared/Controllers/AbortedRpcExceptionFilter.cs` line 9 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeGs7_4E6sBMU07drB:-NdeGs7_4E6sBMU07drC:bq6cymm) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Shared/Controllers/AbortedRpcExceptionFilter.cs#L9)):*
> ```Smalltalk
>         if (context.Exception is RpcException rpcException && rpcException.StatusCode == StatusCode.Aborted)
>         {
>             context.Result = new StatusCodeResult(StatusCodes.Status409Conflict);
> ```

Can we pass a message through from the RpcException?  What are specific examples of this - I couldn't find any in machine...

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 641 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_n4w10Qz_DTpsRzG8:-NdeMEN9B9_09kSg_pq3:bbzq6gm) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L641)):*

We could add a property to the engine model that indicates if pretranslations are available. We would need to update the gRPC API to allow engines to communicate this. I don't think it is really worth the effort right now.
</blockquote></details>

Then, should we wait to pull out the "IsBuilt" functionality until we can move it to the right place?

johnml1135 (commented):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/113)* status: all files reviewed, 7 unresolved discussions (waiting on @ddaspit)

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/113)* status: 5 of 9 files reviewed, 7 unresolved discussions (waiting on @johnml1135)

___
*[`src/Serval.Shared/Controllers/AbortedRpcExceptionFilter.cs` line 9 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeGs7_4E6sBMU07drB:-Ndf8iWn5P69iDheqMN9:b-f3f2i1) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Shared/Controllers/AbortedRpcExceptionFilter.cs#L9)):*

Can we pass a message through from the RpcException?  What are specific examples of this - I couldn't find any in machine...
</blockquote></details>

I passed the message from the RpcException. There aren't any examples yet. We need to implement the checks in Machine.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 612 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-NdeGeITCj2nUajpiAcT:-NdfFbm31oDnJ2FibhoS:b-22lw61) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L612)):*

If we are going to remove this comment here - we should remove it everywhere.
</blockquote></details>

The other methods can still potentially return a 405 or 409.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 641 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_n4w10Qz_DTpsRzG8:-NdfFHWwC2OO_2mtuSiZ:b-8su1ne) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L641)):*

Then, should we wait to pull out the "IsBuilt" functionality until we can move it to the right place?
</blockquote></details>

I added a check for the `ModelRevision` property instead. This property gets incremented on every successful build.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-Ndf96HZ1soc-f41F_dw:b-xj4kbv) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

Combining the 2 functions into 1 - I like it.  Is it truly backwards compatible though?  Wouldn't the syntax of the url be different?  I believe that Peter is already using the "get by textId" function...
</blockquote></details>

No, it isn't backwards compatible. We should let them know that we made a breaking change to the API.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 593 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeG6vZ9ZBrQWyEgLUX:-NdfAjDV623tOf3QkTZC:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L593)):*

...text id if provided.
</blockquote></details>

Done.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 597 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeG1TR82VLD4q0gPFc:-NdfAig25I41Hctqi751:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L597)):*

The text (optional).
</blockquote></details>

Done.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 614 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeFqlT0WVQsB2wfhC8:-NdfEasH12CmslXcATxG:b-5k9q70) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L614)):*

Should this be defaulted to null?  I don't know if it is automatic or not - but it should be explicit...
</blockquote></details>

Defaulting to null doesn't really do anything in this context. This method is called by the ASP.NET Core middleware, so it will pass null if the request url does not have `textId` in the query string.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 614 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeFqlT0WVQsB2wfhC8:-NdfUuXGExcW91ffWky9:bxs1con) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L614)):*

Defaulting to null doesn't really do anything in this context. This method is called by the ASP.NET Core middleware, so it will pass null if the request url does not have `textId` in the query string.
</blockquote></details>

I thought as much - just thought I would ask.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 641 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_n4w10Qz_DTpsRzG8:-NdfV6Ug6fu1rcAR4ufT:b-k6advk) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L641)):*

I added a check for the `ModelRevision` property instead. This property gets incremented on every successful build.
</blockquote></details>

Will this break echo?  Will this work for Nmt and SmtTransfer?

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-NdfVHQ5B3wsR628vVYr:b-y558v6) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

No, it isn't backwards compatible. We should let them know that we made a breaking change to the API.
</blockquote></details>

Hmm... we should probably rev the APi version, keep the old method (calling the new query) and call the old method depreciated.  I would hate to break the current implementation of SF without giving them a chance to fix it.  https://github.com/dotnet/aspnet-api-versioning/wiki/Deprecating-a-Service-Version

johnml1135:
*[`src/Serval.Shared/Controllers/AbortedRpcExceptionFilter.cs` line 9 at r2](https://reviewable.io/reviews/sillsdev/serval/113#-NdeGs7_4E6sBMU07drB:-NdfWWSB7GtvciMpLyug:bpje7lt) ([raw file](https://github.com/sillsdev/serval/blob/372bca8ec78981c6b71f32b6d570b021146ff6e2/src/Serval.Shared/Controllers/AbortedRpcExceptionFilter.cs#L9)):*

I passed the message from the RpcException. There aren't any examples yet. We need to implement the checks in Machine.
</blockquote></details>

Good enough for now.

johnml1135 (changes requested):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/113)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit)

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/113)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135)

___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 641 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_n4w10Qz_DTpsRzG8:-NdfjLm53WYlkv6bzJ70:blrrrkf) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L641)):*

Will this break echo?  Will this work for Nmt and SmtTransfer?
</blockquote></details>

Should work with all engines.
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-NdfjbFADrQ8EAEt6t6M:bcoh8fw) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

Hmm... we should probably rev the APi version, keep the old method (calling the new query) and call the old method depreciated.  I would hate to break the current implementation of SF without giving them a chance to fix it.  https://github.com/dotnet/aspnet-api-versioning/wiki/Deprecating-a-Service-Version
</blockquote></details>

I would rather not go through the trouble. We haven't even released yet. It is literally a [single character code change](https://github.com/sillsdev/web-xforge/blob/82964d03391e7b5574521415cfb0a1264a9871bb/src/SIL.XForge.Scripture/Services/PreTranslationService.cs#L59) for them.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-NdgZDk97_VWH1HlMB4f:b-j4cisi) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

I would rather not go through the trouble. We haven't even released yet. It is literally a [single character code change](https://github.com/sillsdev/web-xforge/blob/82964d03391e7b5574521415cfb0a1264a9871bb/src/SIL.XForge.Scripture/Services/PreTranslationService.cs#L59) for them.
</blockquote></details>

Let's get the sign-off from Peter first.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 658 at r1](https://reviewable.io/reviews/sillsdev/serval/113#-Nd_kl9m5V3nTFHL8lRj:-NdpxvbB6rMXn4qXmcEH:bcnjcb7) ([raw file](https://github.com/sillsdev/serval/blob/1f7e92365403c53f1d75632dee00089b00d41e86/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L658)):*

Let's get the sign-off from Peter first.
</blockquote></details>

We got the sign-off.  Ok.

johnml1135 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/113)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

PR #123, opened 2023-09-06
Poll for cancellation client helper

Enkidu93:
As you requested, John - a temporary fix to avoid adding arbitrary delay.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/123)

### DIFF
--- a/tests/Serval.E2ETests/MissingServicesTests.cs
+++ b/tests/Serval.E2ETests/MissingServicesTests.cs
@@ -9,7 +9,7 @@ namespace Serval.E2ETests
         [SetUp]
         public void Setup()
         {
-            _helperClient = InitializeClient();
+            _helperClient = new ServalClientHelper("https://machine.sil.org", ignoreSSLErrors: true);
         }
 
         [Test]
@@ -112,46 +112,5 @@ namespace Serval.E2ETests
 
         [TearDown]
         public void TearDown() { }
-
-        private ServalClientHelper InitializeClient()
-        {
-            var hostUrl = Environment.GetEnvironmentVariable("SERVAL_HOST_URL");
-            var clientId = Environment.GetEnvironmentVariable("SERVAL_CLIENT_ID");
-            var clientSecret = Environment.GetEnvironmentVariable("SERVAL_CLIENT_SECRET");
-            var authUrl = Environment.GetEnvironmentVariable("SERVAL_AUTH_URL");
-            if (hostUrl == null)
-            {
-                Console.WriteLine(
-                    "You need a serval host url in the environment variable SERVAL_HOST_URL!  Look at README for instructions on getting one."
-                );
-            }
-            else if (clientId == null)
-            {
-                Console.WriteLine(
-                    "You need an auth0 client_id in the environment variable SERVAL_CLIENT_ID!  Look at README for instructions on getting one."
-                );
-            }
-            else if (clientSecret == null)
-            {
-                Console.WriteLine(
-                    "You need an auth0 client_secret in the environment variable SERVAL_CLIENT_SECRET!  Look at README for instructions on getting one."
-                );
-            }
-            else if (authUrl == null)
-            {
-                Console.WriteLine(
-                    "You need an auth0 authorization url in the environment variable SERVAL_AUTH_URL!  Look at README for instructions on getting one."
-                );
-            }
-
-            return new ServalClientHelper(
-                hostUrl,
-                authUrl,
-                "https://machine.sil.org",
-                clientId,
-                clientSecret,
-                ignoreSSLErrors: true
-            );
-        }
     }
 }
--- a/tests/Serval.E2ETests/ServalApiTests.cs
+++ b/tests/Serval.E2ETests/ServalApiTests.cs
@@ -9,48 +9,7 @@ public class ServalApiTests
     [SetUp]
     public void SetUp()
     {
-        _helperClient = InitializeClient();
-    }
-
-    private ServalClientHelper InitializeClient()
-    {
-        var hostUrl = Environment.GetEnvironmentVariable("SERVAL_HOST_URL");
-        var clientId = Environment.GetEnvironmentVariable("SERVAL_CLIENT_ID");
-        var clientSecret = Environment.GetEnvironmentVariable("SERVAL_CLIENT_SECRET");
-        var authUrl = Environment.GetEnvironmentVariable("SERVAL_AUTH_URL");
-        if (hostUrl == null)
-        {
-            Console.WriteLine(
-                "You need a serval host url in the environment variable SERVAL_HOST_URL!  Look at README for instructions on getting one."
-            );
-        }
-        else if (clientId == null)
-        {
-            Console.WriteLine(
-                "You need an auth0 client_id in the environment variable SERVAL_CLIENT_ID!  Look at README for instructions on getting one."
-            );
-        }
-        else if (clientSecret == null)
-        {
-            Console.WriteLine(
-                "You need an auth0 client_secret in the environment variable SERVAL_CLIENT_SECRET!  Look at README for instructions on getting one."
-            );
-        }
-        else if (authUrl == null)
-        {
-            Console.WriteLine(
-                "You need an auth0 authorization url in the environment variable SERVAL_AUTH_URL!  Look at README for instructions on getting one."
-            );
-        }
-
-        return new ServalClientHelper(
-            hostUrl,
-            authUrl,
-            "https://machine.sil.org",
-            clientId,
-            clientSecret,
-            ignoreSSLErrors: true
-        );
+        _helperClient = new ServalClientHelper("https://machine.sil.org", ignoreSSLErrors: true);
     }
 
     [Test]
@@ -166,6 +125,7 @@ public class ServalApiTests
         await _helperClient.AddTextCorpusToEngine(engineId, books, "es", "en", false);
         var cId = await _helperClient.AddTextCorpusToEngine(engineId, new string[] { "3JN.txt" }, "es", "en", true);
         await _helperClient.BuildEngine(engineId);
+        await Task.Delay(1000);
         IList<Pretranslation> lTrans = await _helperClient.translationEnginesClient.GetAllPretranslationsAsync(
             engineId,
             cId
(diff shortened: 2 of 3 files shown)


ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/123)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @Enkidu93 and @johnml1135)

___
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-NdgLDCv8UNDIvzh4c5Y:bm2mz4r) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*
> ```Smalltalk
>         await translationEnginesClient.CancelBuildAsync(engineId);
>         int pollIntervalMs = 500; // start throttle at 0.5 seconds
>         while (true)
> ```

It would be better to do long polling here.

johnml1135:
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-NdgdqWq9l3kiMQBKkyo:bb5vjbh) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*

It would be better to do long polling here.
</blockquote></details>

How do we do long polling waiting for a cancel state?

johnml1135:
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-NdgwGl58QyQJPYq55NU:b-rvw6i5) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*

How do we do long polling waiting for a cancel state?
</blockquote></details>

I'm going to make a small change - we can fix it more later.  Add a max timeout to enable the other tests to go.

johnml1135 (approved):
> Reviewed 3 of 3 files at r1, 3 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/123)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

johnml1135 (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/123)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/123)* status: all files reviewed, 1 unresolved discussion (waiting on @johnml1135)

___
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-Ndk7wsH-chn9dmEKgdA:bwdton2) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*

I'm going to make a small change - we can fix it more later.  Add a max timeout to enable the other tests to go.
</blockquote></details>

```
while (true)
{
    try
    {
        TranslationBuild build = await translationEnginesClient.GetBuildAsync(engineId, buildId, buildRevision + 1);
        if (build.State != JobState.Pending && build.State != JobState.Active)
        {
            break;
        }
        buildRevision = build.Revision;
    }
    catch (ServalApiException ex) when (ex.StatusCode == 408)
    {
        // timed out
    }
}
```

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/123)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93 and @johnml1135)

___
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-NdkQ-9V-MG78ZBzG9ZU:bfm8s0o) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*

^ This definitely seems cleaner if it works. I'm still shaky on the idea behind the build revisions (I think the term is just semantically unusual to me in this context), and I know we've avoided relying on it elsewhere (see `BuildEngine`). What were the issues there again? Wasn't it not updated properly and leading to immediate timeouts? Why is it good to use it here and not there? (Just checking).
</blockquote></details>

I'm not sure why it isn't used in the `BuildEngine` method.

The revision is an int that gets incremented every time the build model is updated. If you ask for a future revision, the request will block until the build is updated to that revision or it times out (40s). If the model is already at or past that revision, then it will return right away with the latest revision. This is referred to as long polling. It is a way of doing a "push" notification using a normal REST API.

Enkidu93:
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-NdkTJFS-tPD5JfnaQGI:bf77x41) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*

I'm not sure why it isn't used in the `BuildEngine` method.

The revision is an int that gets incremented every time the build model is updated. If you ask for a future revision, the request will block until the build is updated to that revision or it times out (40s). If the model is already at or past that revision, then it will return right away with the latest revision. This is referred to as long polling. It is a way of doing a "push" notification using a normal REST API.
</blockquote></details>

I see. I remember now. The issue we had was that the build was not updating its revision regularly enough, so it would always timeout. But I suppose that shouldn't be an issue here as long as the cancellation process properly updates the revision.  

@johnml1135 Does that work? Do you want me to copy-paste this in (ahem...I mean re-implement haha)?

johnml1135:
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-Ndkgmkv7txduI_ZsY5D:b-ynaenj) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*

I see. I remember now. The issue we had was that the build was not updating its revision regularly enough, so it would always timeout. But I suppose that shouldn't be an issue here as long as the cancellation process properly updates the revision.  

@johnml1135 Does that work? Do you want me to copy-paste this in (ahem...I mean re-implement haha)?
</blockquote></details>

I am inclined to leave it as it is right now - which should work.  We will be implementing a Cancelling state in a short while and can fix it better.  @ddaspit if you are convinced that the other way would work better, then we can do it.  Are we sure that revision+1 would catch a cancelled job, even if it did not increase the revision?

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/123)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

___
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-Ndkn6g-EBmBAC4QTs10:bvbyrji) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*

I am inclined to leave it as it is right now - which should work.  We will be implementing a Cancelling state in a short while and can fix it better.  @ddaspit if you are convinced that the other way would work better, then we can do it.  Are we sure that revision+1 would catch a cancelled job, even if it did not increase the revision?
</blockquote></details>

The revision is guaranteed to increment every time you update a model. I would like to be testing the long polling, since it is the intended way to do this kind of thing. It is also the way that SF is using the API. We don't need to do it right now, but eventually we need to refactor this code to use long polling.

johnml1135:
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-NdkplQDCmc25TqT4dqv:bmfh8rm) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*

The revision is guaranteed to increment every time you update a model. I would like to be testing the long polling, since it is the intended way to do this kind of thing. It is also the way that SF is using the API. We don't need to do it right now, but eventually we need to refactor this code to use long polling.
</blockquote></details>

ok.  I'll merge it in and we can fix it better later when we have the Cancelling state.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/123)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

___
*[`tests/Serval.E2ETests/ServalClientHelper.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/serval/123#-NdgLDCv8UNDIvzh4c5X:-NdkqAmM5NZARoRjV16E:bnyqo14) ([raw file](https://github.com/sillsdev/serval/blob/402f160b550e19e329dbe6506d4ed64b9463eb33/tests/Serval.E2ETests/ServalClientHelper.cs#L114)):*

ok.  I'll merge it in and we can fix it better later when we have the Cancelling state.
</blockquote></details>

Just for clarification, the Cancelling state isn't necessary to make long polling work. It should work just fine now.

PR #126, opened 2023-09-11
Serval-side solution to https://github.com/sillsdev/machine/issues/81

Enkidu93:
If you'd prefer, an exception filter could be used rather than try-catch in the necessary endpoints

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/126)

### DIFF
--- a/src/Serval.Shared/Controllers/FailedPreconditionRpcExceptionFilter.cs
+++ b/src/Serval.Shared/Controllers/FailedPreconditionRpcExceptionFilter.cs
@@ -1,0 +1,13 @@
+namespace Serval.Shared.Controllers;
+
+public class FailedPreconditionRpcExceptionFilter : ExceptionFilterAttribute
+{
+    public override void OnException(ExceptionContext context)
+    {
+        if (context.Exception is RpcException rpcException && rpcException.StatusCode == StatusCode.FailedPrecondition)
+        {
+            context.Result = new ConflictObjectResult(rpcException.Message);
+            context.ExceptionHandled = true;
+        }
+    }
+}
--- a/src/Serval.Shared/Controllers/ServalControllerBase.cs
+++ b/src/Serval.Shared/Controllers/ServalControllerBase.cs
@@ -7,6 +7,7 @@
 [TypeFilter(typeof(ServiceUnavailableExceptionFilter))]
 [TypeFilter(typeof(ErrorResultFilter))]
 [TypeFilter(typeof(AbortedRpcExceptionFilter))]
+[TypeFilter(typeof(FailedPreconditionRpcExceptionFilter))]
 public abstract class ServalControllerBase : Controller
 {
     private readonly IAuthorizationService _authService;
--- a/src/Serval.Translation/Controllers/TranslationEnginesController.cs
+++ b/src/Serval.Translation/Controllers/TranslationEnginesController.cs
@@ -216,7 +216,17 @@ public class TranslationEnginesController : ServalControllerBase
         if (!(await AuthorizeAsync(id, cancellationToken)).IsSuccess(out ActionResult? errorResult))
             return errorResult;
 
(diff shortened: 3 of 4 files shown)


johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 227 at r1](https://reviewable.io/reviews/sillsdev/serval/126#-Ne3Q6KN7-KJ7uuDtqnI:-Ne3Q6KN7-KJ7uuDtqnJ:brrbm6l) ([raw file](https://github.com/sillsdev/serval/blob/5f68270d331b76527d6909d4e55f65b398b21615/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L227)):*
> ```Smalltalk
>         {
>             if (r.StatusCode == Grpc.Core.StatusCode.FailedPrecondition)
>                 return Conflict();
> ```

(after it moves to an exception filter) - would it make sense to send up the message in the conflict?  There may be other preconditions - we could tell the user about the specific one (not enough corpora, not built, etc.).

johnml1135 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/126)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

Enkidu93 (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/126#-:-Ne4P0P37cIQZ4kh0pYv:b-896fix)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @johnml1135)

___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 227 at r1](https://reviewable.io/reviews/sillsdev/serval/126#-Ne3Q6KN7-KJ7uuDtqnI:-Ne4P-RH8wZzfGmVYYaJ:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/5f68270d331b76527d6909d4e55f65b398b21615/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L227)):*

(after it moves to an exception filter) - would it make sense to send up the message in the conflict?  There may be other preconditions - we could tell the user about the specific one (not enough corpora, not built, etc.).
</blockquote></details>

Done.

Enkidu93:
> I already added the `AbortedRpcExceptionFilter` for this purpose. The `Aborted` gRPC status code is the equivalent of the 409 HTTP status code. This PR isn't necessary.
> 
> > Reviewed all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/serval/126#-:-Ne4X14k7ae3Ty5bZFf5:bie5tjd)_ status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @johnml1135)

OK, I wondered about that, but `Aborted` seems semantically unlike what a 409 would usually mean. I mean, it would be aborted, but I suppose a lot of things could fall into that category. I'll switch it from `PreconditionFailed`.

ddaspit (commented):
I agree. It is weird. I was trying to decide between `FailedPrecondition` and `Aborted`. At first, I leaned towards `FailedPrecondition`, but I found [Google's canonical error codes for gRPC APIs](https://github.com/googleapis/api-common-protos/blob/main/google/rpc/code.proto), and it states that the HTTP status code 409 maps to `Aborted`, so I went with that. `FailedPrecondition` maps to 400. Really, I wish there were a more appropriate HTTP status code for this situation. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/126#-:-Ne4eH214g0ry5HsBvk0:b-vgpovv)* status: 1 of 4 files reviewed, 1 unresolved discussion

johnml1135:
I am ok with whichever direction you two decide.  It would be nice to have a meaningful message about "failed preconditions" that gets fed up, no matter the actual exception name.

PR #142, opened 2023-09-25
Strict parsing

Enkidu93:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/142)

### DIFF
--- a/src/Serval.Client/Client.g.cs
+++ b/src/Serval.Client/Client.g.cs
@@ -4317,6 +4317,9 @@ namespace Serval.Client
         [Newtonsoft.Json.JsonProperty("dateFinished", Required = Newtonsoft.Json.Required.Default, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
         public System.DateTimeOffset? DateFinished { get; set; } = default!;
 
+        [Newtonsoft.Json.JsonProperty("options", Required = Newtonsoft.Json.Required.Default, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
+        public string? Options { get; set; } = default!;
+
     }
 
     [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "13.18.2.0 (NJsonSchema v10.8.0.0 (Newtonsoft.Json v13.0.0.0))")]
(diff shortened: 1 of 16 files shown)


johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 792 at r1](https://reviewable.io/reviews/sillsdev/serval/142#-NfHI8cU8CllWZh3IbDm:-NfHI8cU8CllWZh3IbDn:bu7nolb) ([raw file](https://github.com/sillsdev/serval/blob/d83c87301d43466d920a36447f6187fa56f95eba/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L792)):*
> ```Smalltalk
>         {
>             return BadRequest(ae.Message);
>         }
> ```

There should really be an explanation for options in the swagger docs.  Possibly something like this:
general:
* options are a json file that changes how the engine is built.
  * General options:
    * strictPasring: throw an error if the corpora are not properly formatted
  * There are no SMT specific options at this time
  * There are no NMT specific options at this time

johnml1135 (changes requested):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/142)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit and @Enkidu93)

ddaspit (changes requested):
I think there was a misunderstanding about the strict parsing flag. I did not intend for the flag to be configured by the client. What I meant is that it would always be set to `true` for Serval but would be set to `false` by default, so that applications using Machine as a library would continue to get the same behavior.

> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/142#-:-NfHM5bT6r8oTutYkEIx:bgfik1c)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93)

___
*[`src/Serval.Translation/Services/EngineService.cs` line 200 at r1](https://reviewable.io/reviews/sillsdev/serval/142#-NfHAgmA70z5bdpulbfx:-NfHAgmA70z5bdpulbfy:b-m8dlv2) ([raw file](https://github.com/sillsdev/serval/blob/d83c87301d43466d920a36447f6187fa56f95eba/src/Serval.Translation/Services/EngineService.cs#L200)):*
> ```Smalltalk
>                 EngineId = engine.Id,
>                 BuildId = build.Id,
>                 Options = build.Options ?? "",
> ```

`Options` is optional, so you shouldn't set it if it is `null`.

PR #231, opened 2023-12-05
Fixes #208

### LINKED ISSUE #208: Which versification is used?
In different versifications, some portions of scripture are in different books.  For instance Bell and the Dragon may be a separate book or part of Daniel.  If there are different versifications, we need to choose a specific versification to apply.   This brings us to our two questions:
* Which versification is used in the pretranslations received
* Which versification is used for filtering books (and potentially chapters)
* What is versification - let's document it.

### Versification for pretranslations
  Part of the magic of Serval is that the clients don't have to think about versification.  To enable this, the versification of the pretranslations should really be the target versification, as per the target Paratext project (even if it is empty).  That is, all the labels applied should be rewritten to the target versification.
* As of note, it may be worthwhile to add another variable when getting the pretranslations to choose one of three versificaitons: target; source; original, and default to target.

### Versification of filtering
 While there is usually only one pair of Paratext projects used now, in the future there may be multiple diverse versions of the Bible used for training.  Therefore, the most straightforward way of filtering would be to use original versification - and document it and be explicit about it.

### Document versification
  This will probably be a new wiki page and some links to it.  If there is something that already exists, let's just use that.  The only existing things found online are:
* https://github.com/sillsdev/scripture
* https://github.com/sillsdev/libpalaso
Neither of which really describe what versification is, the main contours of it, what "original versification" actually means, etc.  Possibly the Wiki for this should end up on libpalaso's wiki.

Enkidu93:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/231)

### DIFF
--- a/src/Serval.Client/Client.g.cs
+++ b/src/Serval.Client/Client.g.cs
@@ -1014,6 +1014,8 @@ namespace Serval.Client
         /// <br/>in the textIds parameter. If the engine does not support pretranslation, these fields have no effect.
         /// <br/>            
         /// <br/>Similarly, specify the corpora and textIds to train on. If no train_on field is provided, all corpora will be used.
+        /// <br/>Paratext projects can be filtered by book for training and pretranslating. This filtering follows the original versification.
+        /// <br/>To filter, use the 3 character code for the book of the Bible in the textID while building. See [here](https://github.com/sillsdev/serval/wiki/Versification-in-Serval) for more information.
         /// <br/>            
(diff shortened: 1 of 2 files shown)


johnml1135 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/231)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Issue #281, opened 2024-01-18
Remove "/translation/engines/queues" endpoint


ddaspit:
There are a couple of issues with this endpoint. It is using POST instead of GET. It should not be under "/translation/engines". We also want to add support for specifying the queue. No one is currently using this endpoint, so we should remove it now, before anyone does.

PR #177, opened 2023-10-11
Queue depth endpoint and status message

Enkidu93:
See [here](https://github.com/sillsdev/machine/pull/110) for parallel PR in machine.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/177)

PR #283, opened 2024-01-22
Add Language Info endpoint

johnml1135:
Move queue endpoint

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)

### DIFF
--- a/samples/EchoTranslationEngine/TranslationEngineServiceV1.cs
+++ b/samples/EchoTranslationEngine/TranslationEngineServiceV1.cs
@@ -292,6 +292,16 @@ public class TranslationEngineServiceV1(BackgroundTaskQueue taskQueue, HealthChe
         return Task.FromResult(new GetQueueSizeResponse { Size = 0 });
     }
 
+    public override Task<GetLanguageInfoResponse> GetLanguageInfo(
+        GetLanguageInfoRequest request,
+        ServerCallContext context
+    )
+    {
+        return Task.FromResult(
+            new GetLanguageInfoResponse { InternalCode = request.Language + "_echo", IsNative = false, }
+        );
+    }
+
     public override async Task<HealthCheckResponse> HealthCheck(Empty request, ServerCallContext context)
     {
         HealthReport healthReport = await _healthCheckService.CheckHealthAsync();
--- a/src/Serval.Client/Client.g.cs
+++ b/src/Serval.Client/Client.g.cs
@@ -1361,24 +1361,24 @@ namespace Serval.Client
         /// <br/>  * The name does not have to be unique, as the engine is uniquely identified by the auto-generated id
         /// <br/>* **sourceLanguage**: The source language code (a valid [IETF language tag](https://en.wikipedia.org/wiki/IETF_language_tag) is recommended)
         /// <br/>* **targetLanguage**: The target language code (a valid IETF language tag is recommended)
-        /// <br/>* **type**: **SmtTransfer** or **Nmt** or **Echo**
-        /// <br/>### SmtTransfer
+        /// <br/>* **type**: **smt-transfer** or **nmt** or **echo**
+        /// <br/>### smt-transfer
         /// <br/>The Statistical Machine Translation Transfer Learning engine is primarily used for translation suggestions. Typical endpoints: translate, get-word-graph, train-segment
-        /// <br/>### Nmt
+        /// <br/>### nmt
         /// <br/>The Neural Machine Translation engine is primarily used for pretranslations.  It is fine-tuned from Meta's NLLB-200. Valid IETF language tags provided to Serval will be converted to [NLLB-200 codes](https://github.com/facebookresearch/flores/tree/main/flores200#languages-in-flores-200).  See more about language tag resolution [here](https://github.com/sillsdev/serval/wiki/Language-Tag-Resolution-for-NLLB%E2%80%90200).
         /// <br/>            
         /// <br/>If you use a language among NLLB's supported languages, Serval will utilize everything the NLLB-200 model already knows about that language when translating. If the language you are working with is not among NLLB's supported languages, the language code will have no effect.
         /// <br/>            
         /// <br/>Typical endpoints: pretranslate
-        /// <br/>### Echo
-        /// <br/>The Echo engine has full coverage of all Nmt and SmtTransfer endpoints. Endpoints like create and build return empty responses. Endpoints like translate and get-word-graph echo the sent content back to the user in a format that mocks Nmt or Smt. For example, translating a segment "test" with the Echo engine would yield a translation response with translation "test". This engine is useful for debugging and testing purposes.
+        /// <br/>### echo
+        /// <br/>The echo engine has full coverage of all nmt and smt-transfer endpoints. Endpoints like create and build return empty responses. Endpoints like translate and get-word-graph echo the sent content back to the user in a format that mocks nmt or Smt. For example, translating a segment "test" with the echo engine would yield a translation response with translation "test". This engine is useful for debugging and testing purposes.
         /// <br/>## Sample request:
         /// <br/>            
         /// <br/>    {
         /// <br/>      "name": "myTeam:myProject:myEngine",
         /// <br/>      "sourceLanguage": "el",
         /// <br/>      "targetLanguage": "en",
-        /// <br/>      "type": "Nmt"
+        /// <br/>      "type": "nmt"
         /// <br/>    }
         /// </remarks>
         /// <param name="engineConfig">The translation engine configuration (see above)</param>
@@ -1403,15 +1403,6 @@ namespace Serval.Client
         /// <returns>The engine was successfully deleted</returns>
         /// <exception cref="ServalApiException">A server side error occurred.</exception>
         System.Threading.Tasks.Task DeleteAsync(string id, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
-
-        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
-        /// <summary>
-        /// Get queue information for a given engine type
-        /// </summary>
-        /// <param name="engineType">A valid engine type: SmtTransfer, Nmt, or Echo</param>
-        /// <returns>Queue information for the specified engine type</returns>
-        /// <exception cref="ServalApiException">A server side error occurred.</exception>
-        System.Threading.Tasks.Task<Queue> GetQueueAsync(string engineType, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
 
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         /// <summary>
@@ -1778,24 +1769,24 @@ namespace Serval.Client
         /// <br/>  * The name does not have to be unique, as the engine is uniquely identified by the auto-generated id
         /// <br/>* **sourceLanguage**: The source language code (a valid [IETF language tag](https://en.wikipedia.org/wiki/IETF_language_tag) is recommended)
         /// <br/>* **targetLanguage**: The target language code (a valid IETF language tag is recommended)
-        /// <br/>* **type**: **SmtTransfer** or **Nmt** or **Echo**
-        /// <br/>### SmtTransfer
+        /// <br/>* **type**: **smt-transfer** or **nmt** or **echo**
+        /// <br/>### smt-transfer
         /// <br/>The Statistical Machine Translation Transfer Learning engine is primarily used for translation suggestions. Typical endpoints: translate, get-word-graph, train-segment
-        /// <br/>### Nmt
+        /// <br/>### nmt
         /// <br/>The Neural Machine Translation engine is primarily used for pretranslations.  It is fine-tuned from Meta's NLLB-200. Valid IETF language tags provided to Serval will be converted to [NLLB-200 codes](https://github.com/facebookresearch/flores/tree/main/flores200#languages-in-flores-200).  See more about language tag resolution [here](https://github.com/sillsdev/serval/wiki/Language-Tag-Resolution-for-NLLB%E2%80%90200).
         /// <br/>            
         /// <br/>If you use a language among NLLB's supported languages, Serval will utilize everything the NLLB-200 model already knows about that language when translating. If the language you are working with is not among NLLB's supported languages, the language code will have no effect.
         /// <br/>            
         /// <br/>Typical endpoints: pretranslate
-        /// <br/>### Echo
-        /// <br/>The Echo engine has full coverage of all Nmt and SmtTransfer endpoints. Endpoints like create and build return empty responses. Endpoints like translate and get-word-graph echo the sent content back to the user in a format that mocks Nmt or Smt. For example, translating a segment "test" with the Echo engine would yield a translation response with translation "test". This engine is useful for debugging and testing purposes.
+        /// <br/>### echo
+        /// <br/>The echo engine has full coverage of all nmt and smt-transfer endpoints. Endpoints like create and build return empty responses. Endpoints like translate and get-word-graph echo the sent content back to the user in a format that mocks nmt or Smt. For example, translating a segment "test" with the echo engine would yield a translation response with translation "test". This engine is useful for debugging and testing purposes.
         /// <br/>## Sample request:
         /// <br/>            
         /// <br/>    {
         /// <br/>      "name": "myTeam:myProject:myEngine",
         /// <br/>      "sourceLanguage": "el",
         /// <br/>      "targetLanguage": "en",
-        /// <br/>      "type": "Nmt"
+        /// <br/>      "type": "nmt"
         /// <br/>    }
         /// </remarks>
         /// <param name="engineConfig">The translation engine configuration (see above)</param>
@@ -2079,106 +2070,6 @@ namespace Serval.Client
                         {
                             string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
                             throw new ServalApiException("The engine does not exist and therefore cannot be deleted", status_, responseText_, headers_, null);
-                        }
-                        else
-                        if (status_ == 503)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("A necessary service is currently unavailable. Check `/health` for more details. ", status_, responseText_, headers_, null);
-                        }
-                        else
-                        {
-                            var responseData_ = response_.Content == null ? null : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The HTTP status code of the response was not expected (" + status_ + ").", status_, responseData_, headers_, null);
-                        }
-                    }
-                    finally
-                    {
-                        if (disposeResponse_)
-                            response_.Dispose();
-                    }
-                }
-            }
-            finally
-            {
-                if (disposeClient_)
-                    client_.Dispose();
-            }
-        }
-
-        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
-        /// <summary>
-        /// Get queue information for a given engine type
-        /// </summary>
-        /// <param name="engineType">A valid engine type: SmtTransfer, Nmt, or Echo</param>
-        /// <returns>Queue information for the specified engine type</returns>
-        /// <exception cref="ServalApiException">A server side error occurred.</exception>
-        public virtual async System.Threading.Tasks.Task<Queue> GetQueueAsync(string engineType, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken))
-        {
-            if (engineType == null)
-                throw new System.ArgumentNullException("engineType");
-
-            var client_ = _httpClient;
-            var disposeClient_ = false;
-            try
-            {
-                using (var request_ = new System.Net.Http.HttpRequestMessage())
-                {
-                    var json_ = Newtonsoft.Json.JsonConvert.SerializeObject(engineType, _settings.Value);
-                    var content_ = new System.Net.Http.StringContent(json_);
-                    content_.Headers.ContentType = System.Net.Http.Headers.MediaTypeHeaderValue.Parse("application/json");
-                    request_.Content = content_;
-                    request_.Method = new System.Net.Http.HttpMethod("POST");
-                    request_.Headers.Accept.Add(System.Net.Http.Headers.MediaTypeWithQualityHeaderValue.Parse("application/json"));
-
-                    var urlBuilder_ = new System.Text.StringBuilder();
-                    if (!string.IsNullOrEmpty(BaseUrl)) urlBuilder_.Append(BaseUrl);
-                    // Operation Path: "translation/engines/queues"
-                    urlBuilder_.Append("translation/engines/queues");
-
-                    PrepareRequest(client_, request_, urlBuilder_);
-
-                    var url_ = urlBuilder_.ToString();
-                    request_.RequestUri = new System.Uri(url_, System.UriKind.RelativeOrAbsolute);
-
-                    PrepareRequest(client_, request_, url_);
-
-                    var response_ = await client_.SendAsync(request_, System.Net.Http.HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
-                    var disposeResponse_ = true;
-                    try
-                    {
-                        var headers_ = new System.Collections.Generic.Dictionary<string, System.Collections.Generic.IEnumerable<string>>();
-                        foreach (var item_ in response_.Headers)
-                            headers_[item_.Key] = item_.Value;
-                        if (response_.Content != null && response_.Content.Headers != null)
-                        {
-                            foreach (var item_ in response_.Content.Headers)
-                                headers_[item_.Key] = item_.Value;
-                        }
-
-                        ProcessResponse(client_, response_);
-
-                        var status_ = (int)response_.StatusCode;
-                        if (status_ == 200)
-                        {
-                            var objectResponse_ = await ReadObjectResponseAsync<Queue>(response_, headers_, cancellationToken).ConfigureAwait(false);
-                            if (objectResponse_.Object == null)
-                            {
-                                throw new ServalApiException("Response was null which was not expected.", status_, objectResponse_.Text, headers_, null);
-                            }
-                            return objectResponse_.Object;
-                        }
-                        else
-                        if (status_ == 401)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The client is not authenticated", status_, responseText_, headers_, null);
-                        }
-                        else
-                        if (status_ == 403)
-                        {
-                            string responseText_ = ( response_.Content == null ) ? string.Empty : await response_.Content.ReadAsStringAsync().ConfigureAwait(false);
-                            throw new ServalApiException("The authenticated client cannot perform the operation", status_, responseText_, headers_, null);
                         }
                         else
                         if (status_ == 503)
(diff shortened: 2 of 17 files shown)


Enkidu93 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @johnml1135)

___
*[`samples/EchoTranslationEngine/TranslationEngineServiceV1.cs` line 303 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-NolYNZ0ATJUT3-vTUtp:-NolYNZ0ATJUT3-vTUtq:bi1patf) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/samples/EchoTranslationEngine/TranslationEngineServiceV1.cs#L303)):*
> ```Smalltalk
>             new GetLanguageInfoResponse
>             {
>                 ResolvedLanguageCode = "hello",
> ```

Given that the Echo engine in general just spits back whatever you give in, would it be worth doing something more akin to that here. I get that it's a bit harder to echo this information since it needs to undergo transformation. If we do keep it this way, we should at least make sure this is documented since this is meant to be used for testing and someone might expect that the response language code, for example, is just the input (?).
___
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 1051 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-NolXjlJ47N5UAPQuvoY:-NolXjlJ47N5UAPQuvoZ:b-w6q1vq) ([raw file](https://github.com/sillsdev/serval/blob/04f977f2ecde44c94b9b9306b22a966aa6242167/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L1051)):*
> ```Smalltalk
>     }
> 
>     private QueueDto Map(Queue source) => new() { Size = source.Size, EngineType = source.EngineType };
> ```

Is the goal to ultimately reimplement something very much akin to this but that allows you to specify which queue? Or am I misunderstanding? Is it just that you want to be able to specify the queue on creation? I thought one of the rationales was that this could be used to probe queue loads before adding jobs (which just specifying a queue in `buildOptions` wouldn't allow you to do. If the plan is to rebuild this slightly differently, then I wonder if it's worth just doing that now in the same PR since the code will be relatively similar and not take much effort. If not, we should probably remove the corresponding code on the Machine side (?). 
___
*[`tests/Serval.ApiServer.IntegrationTests/TranslationEngineTests.cs` line 1424 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-NolWRqe6OTtNbxm7gZj:-NolWRqe6OTtNbxm7gZk:b-wd1zxt) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/tests/Serval.ApiServer.IntegrationTests/TranslationEngineTests.cs#L1424)):*
> ```Smalltalk
>         }
> 
>         public TranslationClient CreateTranslationClient(IEnumerable<string>? scope = null)
> ```

Why make a new client creation method instead of extending the previous one?

Enkidu93:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 1051 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-NolXjlJ47N5UAPQuvoY:-NolZAnWBuISFPs99QUL:bxqigfg) ([raw file](https://github.com/sillsdev/serval/blob/04f977f2ecde44c94b9b9306b22a966aa6242167/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L1051)):*

Is the goal to ultimately reimplement something very much akin to this but that allows you to specify which queue? Or am I misunderstanding? Is it just that you want to be able to specify the queue on creation? I thought one of the rationales was that this could be used to probe queue loads before adding jobs (which just specifying a queue in `buildOptions` wouldn't allow you to do. If the plan is to rebuild this slightly differently, then I wonder if it's worth just doing that now in the same PR since the code will be relatively similar and not take much effort. If not, we should probably remove the corresponding code on the Machine side (?).
</blockquote></details>

Oh, I see - nevermind!

Enkidu93:
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 63 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-Nol_C-a1gCdYSAWL1MT:-Nol_C-a1gCdYSAWL1MU:b-w7pc4f) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/src/Serval.Translation/Controllers/TranslationController.cs#L63)):*
> ```Smalltalk
>     [ProducesResponseType(typeof(void), StatusCodes.Status503ServiceUnavailable)]
>     public async Task<ActionResult<LanguageInfoDto>> GetLanguageInfoAsync(
>         [NotNull] string engineType,
> ```

Same as with queueDepth above. See this message from Damien on this topic during the initial queue depth PR: https://github.com/sillsdev/serval/pull/177#pullrequestreview-1674873160. This is why it inadvertently ended up being a POST request.

johnml1135:
*[`samples/EchoTranslationEngine/TranslationEngineServiceV1.cs` line 303 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-NolYNZ0ATJUT3-vTUtp:-Nolb5PB1tnRMZx8mMA2:b-335fhv) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/samples/EchoTranslationEngine/TranslationEngineServiceV1.cs#L303)):*

Given that the Echo engine in general just spits back whatever you give in, would it be worth doing something more akin to that here. I get that it's a bit harder to echo this information since it needs to undergo transformation. If we do keep it this way, we should at least make sure this is documented since this is meant to be used for testing and someone might expect that the response language code, for example, is just the input (?).
</blockquote></details>

Agreed - that makes more sense.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 63 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-Nol_C-a1gCdYSAWL1MT:-NolcNp999LI2usmjHNy:b-gj3d52) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/src/Serval.Translation/Controllers/TranslationController.cs#L63)):*

Same as with queueDepth above. See this message from Damien on this topic during the initial queue depth PR: https://github.com/sillsdev/serval/pull/177#pullrequestreview-1674873160. This is why it inadvertently ended up being a POST request.
</blockquote></details>

Can you copy the comment from that review here?  I don't know which one you are referring to.

johnml1135:
This also resolves https://github.com/sillsdev/serval/issues/281.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 1051 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-NolXjlJ47N5UAPQuvoY:-NolckVq7fCTB6Z67qp_:bjudz33) ([raw file](https://github.com/sillsdev/serval/blob/04f977f2ecde44c94b9b9306b22a966aa6242167/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L1051)):*

I do still wonder if it's worth adding in that queue-specification in this PR if that is the ultimate intent.
</blockquote></details>

I added the reference to https://github.com/sillsdev/serval/issues/281 above.

johnml1135:
*[`tests/Serval.ApiServer.IntegrationTests/TranslationEngineTests.cs` line 1424 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-NolWRqe6OTtNbxm7gZj:-NolcsxK7ikeO-ts6k8m:b5nn4jd) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/tests/Serval.ApiServer.IntegrationTests/TranslationEngineTests.cs#L1424)):*

Why make a new client creation method instead of extending the previous one?
</blockquote></details>

Am I able to?  The queue and the language info are now in a separate file TranslationController.cs because we needed a different endpoint.  I can merge the two files into one file called TranslationController.cs and redo all the "engines" endpoints - what do you  think is best?

Enkidu93:
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 63 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-Nol_C-a1gCdYSAWL1MT:-NolfmNB7BZH-tPGfDcC:b-1vzkcm) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/src/Serval.Translation/Controllers/TranslationController.cs#L63)):*

Can you copy the comment from that review here?  I don't know which one you are referring to.
</blockquote></details>

URLs are often treated as case insensitive. In this case, just the engine type part of the URL will be case sensitive. Unfortunately, we never expected to use the engine types in a URL, so they can't easily be converted to Pascal case. We have a couple of options:

1.  Change the engine type values to kebab case on all endpoints (breaking change). We convert to Pascal case for internal use or switch to using kebab case everywhere.
2.  Change the engine type values to kebab case only on this endpoint. We convert to Pascal case for internal use.
3.  Pass the engine type in the body or query string.
4.  Require that engine types are case insensitive everywhere.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 63 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-Nol_C-a1gCdYSAWL1MT:-NolgWElCqPAeFyuaZGq:b-2sosyg) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/src/Serval.Translation/Controllers/TranslationController.cs#L63)):*

URLs are often treated as case insensitive. In this case, just the engine type part of the URL will be case sensitive. Unfortunately, we never expected to use the engine types in a URL, so they can't easily be converted to Pascal case. We have a couple of options:

1.  Change the engine type values to kebab case on all endpoints (breaking change). We convert to Pascal case for internal use or switch to using kebab case everywhere.
2.  Change the engine type values to kebab case only on this endpoint. We convert to Pascal case for internal use.
3.  Pass the engine type in the body or query string.
4.  Require that engine types are case insensitive everywhere.
</blockquote></details>

@ddaspit - what do you think?  I am inclined to do the case insensitive route...

ddaspit (changes requested):
> Reviewed 11 of 12 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: all files reviewed, 10 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval.Grpc/Protos/serval/translation/v1/engine.proto` line 85 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-NolfsET8p_3Ot7tiX1v:-NolfsET8p_3Ot7tiX1w:b-2wewi1) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Grpc/Protos/serval/translation/v1/engine.proto#L85)):*
> ```proto
> message GetLanguageInfoRequest {
>     string engine_type = 1;
>     string language_code = 2;
> ```

This should be named `language` to be consistent with the rest of the API.
___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8ZA2C3U9KFT25tB9:-Nom8ZA2C3U9KFT25tBA:b-wrtzoc) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L5)):*
> ```Smalltalk
> public class LanguageInfoDto
> {
>     public string ResolvedLanguageCode { get; set; } = default!;
> ```

I think I would prefer a different name. I don't think "resolved" is very clear in this context. Maybe something like `InternalCode`.
___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom96wD90oiFHjAv86C:-Nom96wD90oiFHjAv86D:bndt5cd) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L6)):*
> ```Smalltalk
> {
>     public string ResolvedLanguageCode { get; set; } = default!;
>     public bool NativeLanguageSupport { get; set; } = default!;
> ```

To be consistent with the rest of the API, you should use `IsSupported`.
___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-Nom9_wA86CDahFu83mG:bibk4u1) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*
> ```Smalltalk
>     public string ResolvedLanguageCode { get; set; } = default!;
>     public bool NativeLanguageSupport { get; set; } = default!;
>     public string CommonLanguageName { get; set; } = default!;
> ```

I'm not sure what "common" means in this context. I think simply `Name` should be sufficient.
___
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 63 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-Nol_C-a1gCdYSAWL1MT:-Nom7rgB44Oy0HmNqOQm:btjft73) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/src/Serval.Translation/Controllers/TranslationController.cs#L63)):*

@ddaspit - what do you think?  I am inclined to do the case insensitive route...
</blockquote></details>

I think we should go with option 1 with one modification. We still support the engine type in pascal case for existing endpoints for backwards compatibility. 
___
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom7ccW7_jIwGEaQTxk:-Nom7ccW7_jIwGEaQTxl:bwxmjqq) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Controllers/TranslationController.cs#L6)):*
> ```Smalltalk
> [Route("api/v{version:apiVersion}/translation")]
> [OpenApiTag("Translation Engines")]
> public class TranslationController(IAuthorizationService authService, IEngineService engineService)
> ```

This should be a translation engine types controller. The route would be `api/v{version:apiVersion}/translation/engine-types`.
___
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 57 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8JYU2ZP-0ybcBXnv:-Nom8JYU2ZP-0ybcBXnw:b-xl5cc3) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Controllers/TranslationController.cs#L57)):*
> ```Smalltalk
>     /// <response code="503">A necessary service is currently unavailable. Check `/health` for more details. </response>
>     [Authorize(Scopes.ReadTranslationEngines)]
>     [HttpGet("{engineType}/language-info/{languageCode}")]
> ```

I would rename the route to `languages`.
___
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 64 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9nzB1s5IzXpFx9lA:-Nom9nzB1s5IzXpFx9lB:b-lxf6iv) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Controllers/TranslationController.cs#L64)):*
> ```Smalltalk
>     public async Task<ActionResult<LanguageInfoDto>> GetLanguageInfoAsync(
>         [NotNull] string engineType,
>         [NotNull] string languageCode,
> ```

This should be simply `language`.

johnml1135:
*[`src/Serval.Grpc/Protos/serval/translation/v1/engine.proto` line 85 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-NolfsET8p_3Ot7tiX1v:-NomEqL-CpoWfj2fuvD4:b-5sg3a3) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Grpc/Protos/serval/translation/v1/engine.proto#L85)):*

This should be named `language` to be consistent with the rest of the API.
</blockquote></details>

Replaced in all locations in serval and machine.

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8ZA2C3U9KFT25tB9:-NomEuQq-nZHoxwWMQOA:b-nbl6pe) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L5)):*

I think I would prefer a different name. I don't think "resolved" is very clear in this context. Maybe something like `InternalCode`.
</blockquote></details>

ISOLanguage?  ISO639_3Language?  InternationalLanguage?

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom96wD90oiFHjAv86C:-NomFFU57nNo9lABkzxQ:bpg2g3o) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L6)):*

To be consistent with the rest of the API, you should use `IsSupported`.
</blockquote></details>

All languages are supported - how about `IsSupportedNatively`?

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-NomFQ-i88aWWmpw16iX:b-fhrj5f) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*

I'm not sure what "common" means in this context. I think simply `Name` should be sufficient.
</blockquote></details>

English, etc.  They are from the table here: https://github.com/facebookresearch/flores/blob/main/flores200/README.md

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom7ccW7_jIwGEaQTxk:-NomFsyqESyy2x-mnhY7:b-iofhvy) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Controllers/TranslationController.cs#L6)):*

This should be a translation engine types controller. The route would be `api/v{version:apiVersion}/translation/engine-types`.
</blockquote></details>

Changed and file renamed.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 57 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8JYU2ZP-0ybcBXnv:-NomFw1l7hyPO5bwbG-B:b-1om1bm) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Controllers/TranslationController.cs#L57)):*

I would rename the route to `languages`.
</blockquote></details>

All "codes" renamed.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 64 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9nzB1s5IzXpFx9lA:-NomG2uHAKWDd7jkAZt0:bb7f2mf) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Controllers/TranslationController.cs#L64)):*

This should be simply `language`.
</blockquote></details>

done

Enkidu93:
*[`tests/Serval.ApiServer.IntegrationTests/TranslationEngineTests.cs` line 1424 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-NolWRqe6OTtNbxm7gZj:-NolfdP7BcJnR1wyj9Vo:bc2a6xd) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/tests/Serval.ApiServer.IntegrationTests/TranslationEngineTests.cs#L1424)):*

Am I able to?  The queue and the language info are now in a separate file TranslationController.cs because we needed a different endpoint.  I can merge the two files into one file called TranslationController.cs and redo all the "engines" endpoints - what do you  think is best?
</blockquote></details>

I see. No, I think your refactor makes sense.

Enkidu93:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom96wD90oiFHjAv86C:-NomHzOM0A2WeHraCxGG:b-w04ieg) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L6)):*

All languages are supported - how about `IsSupportedNatively`?
</blockquote></details>

Where is `IsSupported` used elsewhere in the API? And to be clear, this flag just tells the user whether the language is one of NLLB's 200, right? That's what this means?

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom96wD90oiFHjAv86C:-NomJ3Xm5DiCDqnpqAjO:b-qlrh2j) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L6)):*

Where is `IsSupported` used elsewhere in the API? And to be clear, this flag just tells the user whether the language is one of NLLB's 200, right? That's what this means?
</blockquote></details>

Yes, it says if the language is in the NLLB200, and when MADLAD is supported, it will tell if it is in that model natively.  This can be used for determining if a language is supported for source text or if you can do direct inferencing on it.

johnml1135:
*[`tests/Serval.ApiServer.IntegrationTests/TranslationEngineTests.cs` line 1424 at r1](https://reviewable.io/reviews/sillsdev/serval/283#-NolWRqe6OTtNbxm7gZj:-NomJNyqET-1fMwgGcyj:b-jxy7qt) ([raw file](https://github.com/sillsdev/serval/blob/027027a95d0b224097e751e0347f1b98e187a74f/tests/Serval.ApiServer.IntegrationTests/TranslationEngineTests.cs#L1424)):*

I see. No, I think your refactor makes sense.
</blockquote></details>

I will keep the refactor then and rename the file to TranslationEngineTypesController.cs (as per Damien's comment).

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8ZA2C3U9KFT25tB9:-NomJy_O4sZpbK4lyqVl:b-hp9rs6) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L5)):*

ISOLanguage?  ISO639_3Language?  InternationalLanguage?
</blockquote></details>

ISOLanguageCode?

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom96wD90oiFHjAv86C:-NomNfPi0OeMxzsYmmC1:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L6)):*

OK, yes, makes sense. I don't see an issue with something like `IsSupportedNatively` or `IsSupportedByModel`. If Damien just meant by `IsSupported` that you should give it a name with 'Is' in its name, then cool, but I do think just `IsSupported` might be confusing since we *can* support languages for which this is false in a sense - the model just doesn't have native support without fine-tuning. We can clarify in the docs of course, but it would be nice to have that meaning encoded in the name.
</blockquote></details>

Done.

Enkidu93 (approved):
Everything that caught my attention has been resolved :lgtm:

> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283#-:-NomNkHI31idsSseZSXN:b-357rbr)* status: all files reviewed, 8 unresolved discussions (waiting on @ddaspit)

ddaspit (commented):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: all files reviewed, 5 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8ZA2C3U9KFT25tB9:-NombW-u4adw8rSwQQQN:bcm1jjg) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L5)):*

ISOLanguageCode?
</blockquote></details>

`ISOLanguageCode` isn't really appropriate either. This refers to the token that is used by model to represent a language. For example, in the MADLAD-400 model the English token would be `<2en>`.
___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom96wD90oiFHjAv86C:-NomboAV9ddVLdEJtZ8M:b-zh77ad) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L6)):*

Done.
</blockquote></details>

Yes, I was referring to the convention `Is<name>` for boolean variables. I would just go with `IsNative`. It conveys the same meaning, and it is better to be concise, especially in DTOs.
___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-NomcyikAuTVQ6cqTb6C:bhdvyqs) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*

English, etc.  They are from the table here: https://github.com/facebookresearch/flores/blob/main/flores200/README.md
</blockquote></details>

This field is just the language name. In the context of a `LanguageInfo`, all that is needed is `Name` to convey the same thing.
___
*[`src/Serval.Translation/Controllers/TranslationController.cs` line 57 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8JYU2ZP-0ybcBXnv:-NomdIuMBSxHnab0b3jn:b-2fpo61) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Controllers/TranslationController.cs#L57)):*

All "codes" renamed.
</blockquote></details>

Sorry, I was referring to the route, i.e. `{engineType}/languages/{languageCode}`.

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8ZA2C3U9KFT25tB9:-NomhZMy1hcDcErGoibY:bk2fbhq) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L5)):*

`ISOLanguageCode` isn't really appropriate either. This refers to the token that is used by model to represent a language. For example, in the MADLAD-400 model the English token would be `<2en>`.
</blockquote></details>

I'll switch to InternalCode.

Enkidu93 (approved):
> Reviewed 3 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: all files reviewed, 5 unresolved discussions (waiting on @ddaspit and @johnml1135)

Enkidu93:
@ddaspit I see you re-requested my review; was there something in particular you had in mind for me to look at?

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: all files reviewed, 5 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8ZA2C3U9KFT25tB9:-NomiMBh4NQAsC2OXLRt:bs6j4cw) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L5)):*

I'll switch to InternalCode.
</blockquote></details>

I would also make this optional, since there might be engines where this is irrelevant.
___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-NomiTR66UVDaZ1_6_mf:b-5z5uix) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*

This field is just the language name. In the context of a `LanguageInfo`, all that is needed is `Name` to convey the same thing.
</blockquote></details>

I would make this optional.

ddaspit (commented):
I'm not sure how I did that. You don't need to re-review anything.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283#-:-NomigQo4cRkwlE-zuph:bhhekcp)* status: all files reviewed, 5 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8ZA2C3U9KFT25tB9:-Non150S2Rh3Jq_q1LF0:b-8l6aq) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L5)):*

I would also make this optional, since there might be engines where this is irrelevant.
</blockquote></details>

Hmm - then the entire endpoint would be meaningless?  Would Smt engines have native support?

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 6 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom96wD90oiFHjAv86C:-Non631T-v1u_JQVG_P7:b-9c99zb) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L6)):*

Yes, I was referring to the convention `Is<name>` for boolean variables. I would just go with `IsNative`. It conveys the same meaning, and it is better to be concise, especially in DTOs.
</blockquote></details>

IsNative.  I like it.

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-Non7i8n05yDysn0C3Yd:bi8b7i0) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*

I would make this optional.
</blockquote></details>

Can I update to dotnet standard 2.1?  I want nullable types..

ddaspit (commented):
> Reviewed 8 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8ZA2C3U9KFT25tB9:-Non8xWo8NkcnaL1xxBG:b-rodok6) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L5)):*

Hmm - then the entire endpoint would be meaningless?  Would Smt engines have native support?
</blockquote></details>

The main use case for the endpoint is the `IsNative` property. The `InternalCode` is only relevant for pretrained multilingual NMT models. For SMT, I guess everything is native. 
___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-Non9Sf3Do3P5E5qpTFg:b-7x1o51) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*

Can I update to dotnet standard 2.1?  I want nullable types..
</blockquote></details>

What do you want to update to .NET Standard 2.1? This assembly is a .NET 8 DLL, so it already supports nullable types.

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-NonAUGmC2uJVDwqvjaW:bz5dj49) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*

What do you want to update to .NET Standard 2.1? This assembly is a .NET 8 DLL, so it already supports nullable types.
</blockquote></details>

I think I need to update the language support to "latest" and add "nullable" - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version.  I don't need 2.1 - I was wrong.

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-NonBT3UEmAFRsyLeILD:bf45prn) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*

I think I need to update the language support to "latest" and add "nullable" - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version.  I don't need 2.1 - I was wrong.
</blockquote></details>

If we update to C# latest, we get 400+ warnings about nullable types.  What should we do?  Let's have a short meeting tomorrow.

ddaspit (commented):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-NonFsdM6jwFsg1vapX0:bnlcepo) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*

If we update to C# latest, we get 400+ warnings about nullable types.  What should we do?  Let's have a short meeting tomorrow.
</blockquote></details>

I'm confused. This is already a .NET 8 project, so the default C# language version is 12, which supports nullable types.
___
*[`src/Serval.Translation/Controllers/TranslationEngineTypesController.cs` line 64 at r5](https://reviewable.io/reviews/sillsdev/serval/283#-NonFjVgBd8LBXuc23Dr:-NonFjVgBd8LBXuc23Ds:b-h1s66y) ([raw file](https://github.com/sillsdev/serval/blob/780f5fcb9cbc8db811eef63578f14d01d4ec1611/src/Serval.Translation/Controllers/TranslationEngineTypesController.cs#L64)):*
> ```Smalltalk
>     public async Task<ActionResult<LanguageInfoDto>> GetLanguageInfoAsync(
>         [NotNull] string engineType,
>         [NotNull] string languageCode,
> ```

This should be `language` like everything else.

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom8ZA2C3U9KFT25tB9:-NoqJb5sEq2w0nJ4fAlu:bh1y6u8) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L5)):*

The main use case for the endpoint is the `IsNative` property. The `InternalCode` is only relevant for pretrained multilingual NMT models. For SMT, I guess everything is native.
</blockquote></details>

How about make IsNative an enum? IsNative: True|False|N/A

johnml1135:
*[`src/Serval.Translation/Contracts/TranslationInfoDto.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/serval/283#-Nom9_wA86CDahFu83mF:-NoqK-u59yVlazOJsYEZ:bt02ld1) ([raw file](https://github.com/sillsdev/serval/blob/277b4f4ed93b5486820c373c81d27051d1b00848/src/Serval.Translation/Contracts/TranslationInfoDto.cs#L7)):*

I'm confused. This is already a .NET 8 project, so the default C# language version is 12, which supports nullable types.
</blockquote></details>

It's .NET 7.3 - I can show you what I am seeing.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEngineTypesController.cs` line 64 at r5](https://reviewable.io/reviews/sillsdev/serval/283#-NonFjVgBd8LBXuc23Dr:-NorZJgh4g_7LQN8X5up:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/780f5fcb9cbc8db811eef63578f14d01d4ec1611/src/Serval.Translation/Controllers/TranslationEngineTypesController.cs#L64)):*

This should be `language` like everything else.
</blockquote></details>

Done.

ddaspit (commented):
> Reviewed 11 of 11 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: all files reviewed, 7 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval.Translation/Contracts/TranslationEngineConfigDto.cs` line 33 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosT3KgELVvJDfYvV7O:-NosT3KgELVvJDfYvV7P:b-er7gvn) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Contracts/TranslationEngineConfigDto.cs#L33)):*
> ```Smalltalk
>     {
>         get => _type;
>         set { _type = Engine.ToPascalCase(value); }
> ```

It is not the responsibility of DTOs to convert to pascal case. DTOs typically do not contain any logic. The controller is responsible for interpreting the inputs and mapping them to models that the application/service layer can understand.
___
*[`src/Serval.Translation/Models/Engine.cs` line 14 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosVAiWD83VBscvi9Iu:-NosVAiWD83VBscvi9Iv:b-v1qp0n) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/Engine.cs#L14)):*
> ```Smalltalk
>     {
>         get => _type;
>         set { _type = ToPascalCase(value); }
> ```

It is not the responsibility of models to convert to pascal case. Models typically do not contain any logic. The controller is responsible for interpreting the inputs and mapping them to models that the application/service layer can understand.
___
*[`src/Serval.Translation/Models/Engine.cs` line 29 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosUpsgAwSKABT9QQtW:-NosUpsgAwSKABT9QQtX:b1x7ot0) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/Engine.cs#L29)):*
> ```Smalltalk
>     private static partial Regex SnakeCaseRegex();
> 
>     public static string ToPascalCase(string name)
> ```

Use the [CaseExtensions](https://github.com/vad3x/case-extensions) library to perform the conversion.
___
*[`src/Serval.Translation/Models/LanguageInfo.cs` line 9 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosVNkB6JFsHCk0g3aI:-NosVNkB6JFsHCk0g3aJ:b-v1qp0n) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/LanguageInfo.cs#L9)):*
> ```Smalltalk
>     {
>         get => _engineType;
>         set { _engineType = Engine.ToPascalCase(value); }
> ```

It is not the responsibility of models to convert to pascal case. Models typically do not contain any logic. The controller is responsible for interpreting the inputs and mapping them to models that the application/service layer can understand.
___
*[`src/Serval.Translation/Models/Queue.cs` line 7 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosVOnY3EalBN8XFiIB:-NosVOnY3EalBN8XFiIC:b-v1qp0n) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/Queue.cs#L7)):*
> ```Smalltalk
>     public int Size { get; set; } = default;
>     private string _engineType = default!;
>     public string EngineType
> ```

It is not the responsibility of models to convert to pascal case. Models typically do not contain any logic. The controller is responsible for interpreting the inputs and mapping them to models that the application/service layer can understand.
___
*[`src/Serval.Translation/Controllers/TranslationEngineTypesController.cs` line 57 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosU-XD7blVcMhOlq-V:-NosU-XD7blVcMhOlq-W:bnmzuhr) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Controllers/TranslationEngineTypesController.cs#L57)):*
> ```Smalltalk
>     /// * **EngineType**: See above
>     /// * **IsNative**: Whether language is in the base NLLB-200 model as per [this list](https://github.com/facebookresearch/flores/blob/main/nllb_seed/README.md)
>     /// * **InternalCode**: The ISO 639-3 code that the language maps to according to [these rules](https://github.com/sillsdev/serval/wiki/FLORES%E2%80%90200-Language-Code-Resolution-for-NMT-Engine)
> ```

The FLORES-200 codes are not ISO 639-3 codes. They are a combination of ISO 639-3 codes and ISO 15924 codes. They do not conform to any particular standard. Also, this is only applicable to the NMT engine type using NLLB.

johnml1135:
*[`src/Serval.Translation/Controllers/TranslationEngineTypesController.cs` line 57 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosU-XD7blVcMhOlq-V:-NovbSzg-MHwGwJTmBQR:bq5wuc9) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Controllers/TranslationEngineTypesController.cs#L57)):*

The FLORES-200 codes are not ISO 639-3 codes. They are a combination of ISO 639-3 codes and ISO 15924 codes. They do not conform to any particular standard. Also, this is only applicable to the NMT engine type using NLLB.
</blockquote></details>

Updated comment.

johnml1135:
*[`src/Serval.Translation/Models/Engine.cs` line 14 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosVAiWD83VBscvi9Iu:-NovbWR2-Zo_pmBc7mve:b6fih1x) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/Engine.cs#L14)):*

It is not the responsibility of models to convert to pascal case. Models typically do not contain any logic. The controller is responsible for interpreting the inputs and mapping them to models that the application/service layer can understand.
</blockquote></details>

I'll fix all.

johnml1135:
*[`src/Serval.Translation/Models/Engine.cs` line 29 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosUpsgAwSKABT9QQtW:-Novcf4kD_vPtEagpPgC:b-snws76) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/Engine.cs#L29)):*

Use the [CaseExtensions](https://github.com/vad3x/case-extensions) library to perform the conversion.
</blockquote></details>

Done - nice find.

johnml1135:
*[`src/Serval.Translation/Models/LanguageInfo.cs` line 9 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosVNkB6JFsHCk0g3aI:-NovcjsU2ybESA0r6zxf:b8d2z4x) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/LanguageInfo.cs#L9)):*

It is not the responsibility of models to convert to pascal case. Models typically do not contain any logic. The controller is responsible for interpreting the inputs and mapping them to models that the application/service layer can understand.
</blockquote></details>

Will fix all.

johnml1135:
*[`src/Serval.Translation/Models/Queue.cs` line 7 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosVOnY3EalBN8XFiIB:-Novcm4q0ISdzBEo7QrV:b8d2z4x) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/Queue.cs#L7)):*

It is not the responsibility of models to convert to pascal case. Models typically do not contain any logic. The controller is responsible for interpreting the inputs and mapping them to models that the application/service layer can understand.
</blockquote></details>

Will fix all.

ddaspit (commented):
> Reviewed 11 of 11 files at r7, all commit messages.
> Dismissed @Enkidu93 from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval.Translation/Models/LanguageInfo.cs` line 9 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosVNkB6JFsHCk0g3aI:-Now2meG-6_bUA27tyP0:b-3x0ovs) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/LanguageInfo.cs#L9)):*

Will fix all.
</blockquote></details>

I think you accidentally turned this into a field instead of a property.
___
*[`tests/Serval.Translation.Tests/Services/EngineModelTests.cs` line 0 at r7](https://reviewable.io/reviews/sillsdev/serval/283#-Now30fH1XQf379PisCD:-Now30fH1XQf379PisCE:bgqry29) ([raw file](https://github.com/sillsdev/serval/blob/c38f6e877f0cd06a525cc509abbc92338c52f83f/tests/Serval.Translation.Tests/Services/EngineModelTests.cs#L0)):*
We don't need to unit test a library. You can delete this test fixture.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval.Translation/Models/LanguageInfo.cs` line 8 at r7](https://reviewable.io/reviews/sillsdev/serval/283#-Np1j3Gb4hIQ_ppqlL6_:-Np1j3Gb4hIQ_ppqlL6a:b-8x4ma3) ([raw file](https://github.com/sillsdev/serval/blob/c38f6e877f0cd06a525cc509abbc92338c52f83f/src/Serval.Translation/Models/LanguageInfo.cs#L8)):*
> ```Smalltalk
>     public bool IsNative { get; set; } = default!;
>     public string? InternalCode { get; set; } = default!;
>     public string? Name { get; set; } = default!;
> ```

Is there an actual requirement for the language name from Scripture Forge? If not, we should just remove it. We don't need it and it will only make it more difficult to implement this in future engines.

johnml1135:
*[`src/Serval.Translation/Models/LanguageInfo.cs` line 9 at r6](https://reviewable.io/reviews/sillsdev/serval/283#-NosVNkB6JFsHCk0g3aI:-Np6-ERTCxoHqKSQIhDX:b6i4sqo) ([raw file](https://github.com/sillsdev/serval/blob/87ff0d6245a43461bb0b7f8de641be174201069d/src/Serval.Translation/Models/LanguageInfo.cs#L9)):*

I think you accidentally turned this into a field instead of a property.
</blockquote></details>

GAAAAAAAHHHHH!  C# has so many subtle things that have big impacts!  Fixed.

johnml1135:
*[`src/Serval.Translation/Models/LanguageInfo.cs` line 8 at r7](https://reviewable.io/reviews/sillsdev/serval/283#-Np1j3Gb4hIQ_ppqlL6_:-Np60Op-5euSnAZP3bPe:b-lkgsd5) ([raw file](https://github.com/sillsdev/serval/blob/c38f6e877f0cd06a525cc509abbc92338c52f83f/src/Serval.Translation/Models/LanguageInfo.cs#L8)):*

Is there an actual requirement for the language name from Scripture Forge? If not, we should just remove it. We don't need it and it will only make it more difficult to implement this in future engines.
</blockquote></details>

Hmm.  I see what you mean.  I will kill it at the API level but keep the guts of it there.  If we do want it in the future it will be easy to add.  We can also grab data from the ethnologue or similar if needed.

johnml1135:
*[`tests/Serval.Translation.Tests/Services/EngineModelTests.cs` line  at r7](https://reviewable.io/reviews/sillsdev/serval/283#-Now30fH1XQf379PisCD:-Np60df21RmnbV4jbwKL:bbvs8f4) ([raw file](https://github.com/sillsdev/serval/blob/c38f6e877f0cd06a525cc509abbc92338c52f83f/tests/Serval.Translation.Tests/Services/EngineModelTests.cs#L)):*

We don't need to unit test a library. You can delete this test fixture.
</blockquote></details>

Your right - I forgot that I didn't need this anymore.

ddaspit (approved):
:lgtm:

> Reviewed 9 of 9 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283#-:-Np6t0uAEUd-MvIY68IF:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Enkidu93 (approved):
> Reviewed 2 of 11 files at r6, 7 of 11 files at r7, 9 of 9 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/283)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

PR #312, opened 2024-02-12
Update .editorconfig with many changes

johnml1135:
* Enforced naming conventions
* ConfigureAwait for DataAccess (only)
* Always use primary constructor and other .NET 8 conventions
* Fix all warnings from updated editorconfig

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/312)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,5 +8,447 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 4
 
+# avoid this. unless absolutely necessary
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:suggestion
+dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_event = false:suggestion
+
+# Types: use keywords instead of BCL types, and permit var only when the type is clear
+csharp_style_var_for_built_in_types = false:suggestion
+csharp_style_var_when_type_is_apparent = true:suggestion
+csharp_style_var_elsewhere = false:suggestion
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+dotnet_style_predefined_type_for_member_access = true:suggestion
+
+dotnet_sort_system_directives_first = true:warning
+
+# use int x = .. over Int32
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+
+# use int.MaxValue over Int32.MaxValue
+dotnet_style_predefined_type_for_member_access = true:suggestion
+
+# Disallow throw expressions.
+csharp_style_throw_expression = false:warning
+
+# Newline settings
+csharp_new_line_before_open_brace = all
+csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+
+# Namespace settings
+csharp_style_namespace_declarations = file_scoped
+
+# Non-private static fields are PascalCase
+dotnet_naming_rule.non_private_static_fields_should_be_pascal_case.severity = warning
+dotnet_naming_rule.non_private_static_fields_should_be_pascal_case.symbols = non_private_static_fields
+dotnet_naming_rule.non_private_static_fields_should_be_pascal_case.style = non_private_static_field_style
+
+dotnet_naming_symbols.non_private_static_fields.applicable_kinds = field
+dotnet_naming_symbols.non_private_static_fields.applicable_accessibilities = public, protected, internal, protected_internal, private_protected
+dotnet_naming_symbols.non_private_static_fields.required_modifiers = static
+
+dotnet_naming_style.non_private_static_field_style.capitalization = pascal_case
+
+# Non-private readonly fields are PascalCase
+dotnet_naming_rule.non_private_readonly_fields_should_be_pascal_case.severity = warning
+dotnet_naming_rule.non_private_readonly_fields_should_be_pascal_case.symbols = non_private_readonly_fields
+dotnet_naming_rule.non_private_readonly_fields_should_be_pascal_case.style = non_private_readonly_field_style
+
+dotnet_naming_symbols.non_private_readonly_fields.applicable_kinds = field
+dotnet_naming_symbols.non_private_readonly_fields.applicable_accessibilities = public, protected, internal, protected_internal, private_protected
+dotnet_naming_symbols.non_private_readonly_fields.required_modifiers = readonly
+
+dotnet_naming_style.non_private_readonly_field_style.capitalization = pascal_case
+
+# Constants are PascalCase
+dotnet_naming_rule.constants_should_be_pascal_case.severity = warning
+dotnet_naming_rule.constants_should_be_pascal_case.symbols = constants
+dotnet_naming_rule.constants_should_be_pascal_case.style = constant_style
+
+dotnet_naming_symbols.constants.applicable_kinds = field, local
+dotnet_naming_symbols.constants.required_modifiers = const
+
+dotnet_naming_style.constant_style.capitalization = pascal_case
+
+# Private Static fields are PascalCase
+dotnet_naming_rule.static_fields_should_be_camel_case.severity = warning
+dotnet_naming_rule.static_fields_should_be_camel_case.symbols = static_fields
+dotnet_naming_rule.static_fields_should_be_camel_case.style = static_field_style
+
+dotnet_naming_symbols.static_fields.applicable_kinds = field
+dotnet_naming_symbols.static_fields.required_modifiers = static
+
+dotnet_naming_style.static_field_style.capitalization = pascal_case
+
+# Instance fields are camelCase and start with _
+dotnet_naming_rule.instance_fields_should_be_camel_case.severity = warning
+dotnet_naming_rule.instance_fields_should_be_camel_case.symbols = instance_fields
+dotnet_naming_rule.instance_fields_should_be_camel_case.style = instance_field_style
+
+dotnet_naming_symbols.instance_fields.applicable_kinds = field
+
+dotnet_naming_style.instance_field_style.capitalization = camel_case
+dotnet_naming_style.instance_field_style.required_prefix = _
+
+# Locals and parameters are camelCase
+dotnet_naming_rule.locals_should_be_camel_case.severity = warning
+dotnet_naming_rule.locals_should_be_camel_case.symbols = locals_and_parameters
+dotnet_naming_rule.locals_should_be_camel_case.style = camel_case_style
+
+dotnet_naming_symbols.locals_and_parameters.applicable_kinds = parameter, local
+
+dotnet_naming_style.camel_case_style.capitalization = camel_case
+
+# Local functions are PascalCase
+dotnet_naming_rule.local_functions_should_be_pascal_case.severity = warning
+dotnet_naming_rule.local_functions_should_be_pascal_case.symbols = local_functions
+dotnet_naming_rule.local_functions_should_be_pascal_case.style = local_function_style
+
+dotnet_naming_symbols.local_functions.applicable_kinds = local_function
+
+dotnet_naming_style.local_function_style.capitalization = pascal_case
+
+# By default, name items with PascalCase
+dotnet_naming_rule.members_should_be_pascal_case.severity = warning
+dotnet_naming_rule.members_should_be_pascal_case.symbols = all_members
+dotnet_naming_rule.members_should_be_pascal_case.style = pascal_case_style
+
+dotnet_naming_symbols.all_members.applicable_kinds = *
+
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+
+# build checks
+dotnet_diagnostic.IDE0005.severity = error # Remove unnecessary import
+
+# SYSLIB1054: Use 'LibraryImportAttribute' instead of 'DllImportAttribute' to generate P/Invoke marshalling code at compile time
+dotnet_diagnostic.SYSLIB1054.severity = warning
+
+# CA1018: Mark attributes with AttributeUsageAttribute
+dotnet_diagnostic.CA1018.severity = warning
+
+# CA1047: Do not declare protected member in sealed type
+dotnet_diagnostic.CA1047.severity = warning
+
+# CA1305: Specify IFormatProvider
+dotnet_diagnostic.CA1305.severity = warning
+
+# CA1507: Use nameof to express symbol names
+dotnet_diagnostic.CA1507.severity = warning
+
+# CA1708: Identifiers should differ by more than case
+dotnet_diagnostic.CA1708.severity = warning
+
+# CA1510: Use ArgumentNullException throw helper
+dotnet_diagnostic.CA1510.severity = warning
+
+# CA1511: Use ArgumentException throw helper
+dotnet_diagnostic.CA1511.severity = warning
+
+# CA1512: Use ArgumentOutOfRangeException throw helper
+dotnet_diagnostic.CA1512.severity = warning
+
+# CA1513: Use ObjectDisposedException throw helper
+dotnet_diagnostic.CA1513.severity = warning
+
+# CA1715: Identifiers should have correct prefix
+dotnet_diagnostic.CA1715.severity = warning
+
+# CA1720: Identifiers should not contain type names
+dotnet_diagnostic.CA1720.severity = warning
+
+# CA1724: Type Names Should Not Match Namespaces
+dotnet_diagnostic.CA1724.severity = suggestion
+
+# CA1725: Parameter names should match base declaration
+dotnet_diagnostic.CA1725.severity = suggestion
+
+# CA1802: Use literals where appropriate
+dotnet_diagnostic.CA1802.severity = warning
+
+# CA1805: Do not initialize unnecessarily
+dotnet_diagnostic.CA1805.severity = suggestion
+
+# CA1810: Do not initialize unnecessarily
+dotnet_diagnostic.CA1810.severity = warning
+
+# CA1821: Remove empty Finalizers
+dotnet_diagnostic.CA1821.severity = warning
+
+# CA1822: Make member static
+dotnet_diagnostic.CA1822.severity = warning
+dotnet_code_quality.CA1822.api_surface = private, internal
+
+# CA1823: Avoid unused private fields
+dotnet_diagnostic.CA1823.severity = warning
+
+# CA1825: Avoid zero-length array allocations
+dotnet_diagnostic.CA1825.severity = warning
+
+# CA1826: Do not use Enumerable methods on indexable collections. Instead use the collection directly
+dotnet_diagnostic.CA1826.severity = warning
+
+# CA1827: Do not use Count() or LongCount() when Any() can be used
+dotnet_diagnostic.CA1827.severity = warning
+
+# CA1828: Do not use CountAsync() or LongCountAsync() when AnyAsync() can be used
+dotnet_diagnostic.CA1828.severity = warning
+
+# CA1829: Use Length/Count property instead of Count() when available
+dotnet_diagnostic.CA1829.severity = warning
+
+# CA1830: Prefer strongly-typed Append and Insert method overloads on StringBuilder
+dotnet_diagnostic.CA1830.severity = warning
+
+# CA1831: Use AsSpan or AsMemory instead of Range-based indexers when appropriate
+dotnet_diagnostic.CA1831.severity = warning
+
+# CA1832: Use AsSpan or AsMemory instead of Range-based indexers when appropriate
+dotnet_diagnostic.CA1832.severity = warning
+
+# CA1833: Use AsSpan or AsMemory instead of Range-based indexers when appropriate
+dotnet_diagnostic.CA1833.severity = warning
+
+# CA1834: Consider using 'StringBuilder.Append(char)' when applicable
+dotnet_diagnostic.CA1834.severity = warning
+
+# CA1835: Prefer the 'Memory'-based overloads for 'ReadAsync' and 'WriteAsync'
+dotnet_diagnostic.CA1835.severity = warning
+
+# CA1836: Prefer IsEmpty over Count
+dotnet_diagnostic.CA1836.severity = warning
+
+# CA1837: Use 'Environment.ProcessId'
+dotnet_diagnostic.CA1837.severity = warning
+
+# CA1838: Avoid 'StringBuilder' parameters for P/Invokes
+dotnet_diagnostic.CA1838.severity = warning
+
+# CA1839: Use 'Environment.ProcessPath'
+dotnet_diagnostic.CA1839.severity = warning
+
+# CA1840: Use 'Environment.CurrentManagedThreadId'
+dotnet_diagnostic.CA1840.severity = warning
+
+# CA1841: Prefer Dictionary.Contains methods
+dotnet_diagnostic.CA1841.severity = warning
+
+# CA1842: Do not use 'WhenAll' with a single task
+dotnet_diagnostic.CA1842.severity = warning
+
+# CA1843: Do not use 'WaitAll' with a single task
+dotnet_diagnostic.CA1843.severity = warning
+
+# CA1844: Provide memory-based overrides of async methods when subclassing 'Stream'
+dotnet_diagnostic.CA1844.severity = warning
+
+# CA1845: Use span-based 'string.Concat'
+dotnet_diagnostic.CA1845.severity = warning
+
+# CA1846: Prefer AsSpan over Substring
+dotnet_diagnostic.CA1846.severity = warning
+
+# CA1847: Use string.Contains(char) instead of string.Contains(string) with single characters
+dotnet_diagnostic.CA1847.severity = warning
+
+# CA1854: Prefer the IDictionary.TryGetValue(TKey, out TValue) method
+dotnet_diagnostic.CA1854.severity = warning
+
+# CA1855: Prefer 'Clear' over 'Fill'
+dotnet_diagnostic.CA1855.severity = warning
+
+# CA1856: Incorrect usage of ConstantExpected attribute
+dotnet_diagnostic.CA1856.severity = error
+
+# CA1857: A constant is expected for the parameter
+dotnet_diagnostic.CA1857.severity = warning
+
+# CA1858: Use 'StartsWith' instead of 'IndexOf'
+dotnet_diagnostic.CA1858.severity = warning
+
+# CA2008: Do not create tasks without passing a TaskScheduler
+dotnet_diagnostic.CA2008.severity = warning
+
+# CA2009: Do not call ToImmutableCollection on an ImmutableCollection value
+dotnet_diagnostic.CA2009.severity = warning
+
+# CA2011: Avoid infinite recursion
+dotnet_diagnostic.CA2011.severity = warning
+
+# CA2012: Use ValueTask correctly
+dotnet_diagnostic.CA2012.severity = warning
+
+# CA2013: Do not use ReferenceEquals with value types
+dotnet_diagnostic.CA2013.severity = warning
+
+# CA2014: Do not use stackalloc in loops.
+dotnet_diagnostic.CA2014.severity = warning
+
+# CA2016: Forward the 'CancellationToken' parameter to methods that take one
+dotnet_diagnostic.CA2016.severity = warning
+
+# CA2200: Rethrow to preserve stack details
+dotnet_diagnostic.CA2200.severity = warning
+
+# CA2201: Do not raise reserved exception types
+dotnet_diagnostic.CA2201.severity = warning
+
+# CA2208: Instantiate argument exceptions correctly
+dotnet_diagnostic.CA2208.severity = warning
+
+# CA2245: Do not assign a property to itself
+dotnet_diagnostic.CA2245.severity = warning
+
+# CA2246: Assigning symbol and its member in the same statement
+dotnet_diagnostic.CA2246.severity = warning
+
+# CA2249: Use string.Contains instead of string.IndexOf to improve readability.
+dotnet_diagnostic.CA2249.severity = warning
+
+# IDE0005: Remove unnecessary usings
+dotnet_diagnostic.IDE0005.severity = warning
+
+# IDE0011: Curly braces to surround blocks of code
+csharp_prefer_braces = when_multiline : warning
+dotnet_diagnostic.IDE0011.severity = warning
+
+# IDE0020: Use pattern matching to avoid is check followed by a cast (with variable)
+dotnet_diagnostic.IDE0020.severity = warning
+
+# IDE0029: Use coalesce expression (non-nullable types)
+dotnet_diagnostic.IDE0029.severity = warning
+
+# IDE0030: Use coalesce expression (nullable types)
+dotnet_diagnostic.IDE0030.severity = warning
+
+# IDE0031: Use null propagation
+dotnet_diagnostic.IDE0031.severity = warning
+
+# IDE0035: Remove unreachable code
+dotnet_diagnostic.IDE0035.severity = warning
+
+# IDE0036: Order modifiers
+csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:warning
+dotnet_diagnostic.IDE0036.severity = warning
+
+# IDE0038: Use pattern matching to avoid is check followed by a cast (without variable)
+dotnet_diagnostic.IDE0038.severity = warning
+
+# IDE0043: Format string contains invalid placeholder
+dotnet_diagnostic.IDE0043.severity = warning
+
+# IDE0044: Make field readonly
+dotnet_diagnostic.IDE0044.severity = warning
+
+# IDE0051: Remove unused private members
+dotnet_diagnostic.IDE0051.severity = warning
+
+# IDE0055: All formatting rules
+dotnet_diagnostic.IDE0055.severity = suggestion
+
+# IDE0059: Unnecessary assignment to a value
+dotnet_diagnostic.IDE0059.severity = warning
+
+# IDE0060: Remove unused parameter
+dotnet_code_quality_unused_parameters = non_public
+dotnet_diagnostic.IDE0060.severity = warning
+
+# IDE0062: Make local function static
+dotnet_diagnostic.IDE0062.severity = warning
+
+# IDE0074: Use compound assignment
+dotnet_diagnostic.IDE0074.severity = warning
+
+# IDE0161: Convert to file-scoped namespace
+dotnet_diagnostic.IDE0161.severity = warning
+
+# IDE0200: Lambda expression can be removed
+dotnet_diagnostic.IDE0200.severity = warning
+
+# IDE2000: Disallow multiple blank lines
+dotnet_style_allow_multiple_blank_lines_experimental = false
+dotnet_diagnostic.IDE2000.severity = warning
+
+# IDE1006: Naming rule violation
+dotnet_diagnostic.IDE1006.severity = warning
+
+# IDE0270: coalesce null check throw
+dotnet_diagnostic.IDE0270.severity = none
+
+# CA2254: The logging message template should not vary between calls
+dotnet_diagnostic.CA2254.severity = warning
+
+# IDE0007: Use implicit type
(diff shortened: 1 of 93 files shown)


Enkidu93 (changes requested):
Did you get this configuration from somewhere? Was it recommended by someone? Or did you compile these specific configurations yourself?

Also, I imagine Damien will be more opinionated, so I'd definitely wait to get his feedback. I don't have many .NET-specific coding opinions.

> Reviewed 89 of 89 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312#-:-NqXj6sU9ApDYIZQ4XR4:b-r27egj)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/Serval.Grpc/Utils/WriteGrpcHealthCheckResponse.cs` line 18 at r1](https://reviewable.io/reviews/sillsdev/serval/312#-NqXg0G-3H4PkNmaJcp0:-NqXg0G-3H4PkNmaJcp1:b-s3scdz) ([raw file](https://github.com/sillsdev/serval/blob/95e0586e9789dc7082273d591ca694b9ec96f05b/src/Serval.Grpc/Utils/WriteGrpcHealthCheckResponse.cs#L18)):*
> ```Smalltalk
>             if ((entry.Value.Exception?.ToString() ?? "") != "")
>             {
>                 if (healthCheckResultException is null)
> ```

Is this really best practice? I suppose this code is unlikely to be extended very much. I was taught to never use if without braces in c-inspired languages except perhaps in a freestanding, single-line (of course) if-statement - and absolutely never in nested statements. Without braces, it's easy to add cases or more lines of condition-guarded logic incorrectly. The auto-formatting helps, but even so. Like I said, maybe in this function it's not an issue, but I'm not sure I'd like warnings attached to all of my if statements which I might leave with braces intentionally for extensibility/readability. 
___
*[`tests/Serval.ApiServer.IntegrationTests/DataFilesTests.cs` line 70 at r1](https://reviewable.io/reviews/sillsdev/serval/312#-NqXiCKt6NytJwUn6ZE4:-NqXiCKt6NytJwUn6ZE5:b-ykd50g) ([raw file](https://github.com/sillsdev/serval/blob/95e0586e9789dc7082273d591ca694b9ec96f05b/tests/Serval.ApiServer.IntegrationTests/DataFilesTests.cs#L70)):*
> ```Smalltalk
>                 goto case 403;
>             case 403:
>                 ServalApiException? ex = Assert.ThrowsAsync<ServalApiException>(async () =>
> ```

I do think these are fairly clear with 'var', but I suppose it doesn't hurt. Maybe some of these settings should be configured differently for tests?

johnml1135:
I compiled it from editorconfigs from aspdotnet and a few other areas.  It includes:
* ASPDOTNET variable naming
* fairly stringent style guidelines - there is only one way to do it
* No single-line null throwing

The goal is to have the review process take less time overall because these issues are auto-resolved.

johnml1135:
*[`src/Serval.Grpc/Utils/WriteGrpcHealthCheckResponse.cs` line 18 at r1](https://reviewable.io/reviews/sillsdev/serval/312#-NqXg0G-3H4PkNmaJcp0:-NqY4Hxl6l5dMAMlRkjz:bjt31y4) ([raw file](https://github.com/sillsdev/serval/blob/95e0586e9789dc7082273d591ca694b9ec96f05b/src/Serval.Grpc/Utils/WriteGrpcHealthCheckResponse.cs#L18)):*

Is this really best practice? I suppose this code is unlikely to be extended very much. I was taught to never use if without braces in c-inspired languages except perhaps in a freestanding, single-line (of course) if-statement - and absolutely never in nested statements. Without braces, it's easy to add cases or more lines of condition-guarded logic incorrectly. The auto-formatting helps, but even so. Like I said, maybe in this function it's not an issue, but I'm not sure I'd like warnings attached to all of my if statements which I might leave with braces intentionally for extensibility/readability.
</blockquote></details>

The only issue I found with Csharpier formatting has been nested `foreach` statements.  Yes, this is relying on the nesting for clarity and having the code be readable as if it was python.  Currently, we are doing a mix of each throughout our code - this is to encode that yes, you don't have to put in braces in each place.  If you want to for clarity, it is still possible.

johnml1135:
*[`tests/Serval.ApiServer.IntegrationTests/DataFilesTests.cs` line 70 at r1](https://reviewable.io/reviews/sillsdev/serval/312#-NqXiCKt6NytJwUn6ZE4:-NqY5d5q-6_w971UKiYK:bsg4tkr) ([raw file](https://github.com/sillsdev/serval/blob/95e0586e9789dc7082273d591ca694b9ec96f05b/tests/Serval.ApiServer.IntegrationTests/DataFilesTests.cs#L70)):*

I do think these are fairly clear with 'var', but I suppose it doesn't hurt. Maybe some of these settings should be configured differently for tests?
</blockquote></details>

These are the rules:
```
csharp_style_var_for_built_in_types = false:suggestion
csharp_style_var_when_type_is_apparent = true:suggestion
csharp_style_var_elsewhere = false:suggestion
```
There is a specific definition of "is apparent" which is a little different than the normal reading of it.  I have set it so that the use of var is deterministic - and there would be no questions about whether it was appropriate or not.

Enkidu93 (changes requested):
> Reviewed 14 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312)* status: all files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`.editorconfig` line 198 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcgCHfAuzF8tL-VFBE:-NqcgCHfAuzF8tL-VFBF:b-1hxmqg) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L198)):*
> ```EditorConfig
> 
> # CA1826: Do not use Enumerable methods on indexable collections. Instead use the collection directly
> dotnet_diagnostic.CA1826.severity = suggestion
> ```

Why the change to suggestion on this particular one?
___
*[`src/SIL.DataAccess/ExpressionHelper.cs` line 32 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcgYknEIPgkBrDnv38:-NqcgYknEIPgkBrDnv39:b311dwh) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/ExpressionHelper.cs#L32)):*
> ```Smalltalk
>     }
> 
>     public static TV? FindEqualsConstantValue<T, TV>(Expression<Func<T, TV>> field, Expression expression)
> ```

Explain this one to me - what's the idea behind changing the generic name?

johnml1135:
*[`.editorconfig` line 198 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcgCHfAuzF8tL-VFBE:-Nqd0nb7BCBNDoxBKR_N:bhsuqpk) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L198)):*

Why the change to suggestion on this particular one?
</blockquote></details>

This one should be a warning...

johnml1135:
*[`src/SIL.DataAccess/ExpressionHelper.cs` line 32 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcgYknEIPgkBrDnv38:-Nqd1-RdA8oRklJ4SGxx:b-cff9o7) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/ExpressionHelper.cs#L32)):*

Explain this one to me - what's the idea behind changing the generic name?
</blockquote></details>

Renamed to TConstant.  It was complaining about https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1715.

ddaspit (changes requested):
> Reviewed 78 of 89 files at r1, 14 of 14 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312)* status: all files reviewed, 10 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`.editorconfig` line 27 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKvTR8frYU13jAGB2:-NqcKvTR8frYU13jAGB3:b3y8sdy) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L27)):*
> ```EditorConfig
> 
> # Don't use this. qualifier
> dotnet_style_qualification_for_field = false:suggestion
> ```

I think this is already specified elsewhere.
___
*[`.editorconfig` line 82 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcIUlcBb1ZUcXy0rXg:-NqcIUlcBb1ZUcXy0rXh:b-bxyb4j) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L82)):*
> ```EditorConfig
> dotnet_naming_style.constant_style.capitalization = pascal_case
> 
> # Static fields are camelCase and start with s_
> ```

I don't like the `s_` for static fields. Static read-only fields are essentially constants, so I prefer them to use pascal case.
___
*[`.editorconfig` line 173 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd4OMI7d23JrzZVM_-:-Nqd4OMI7d23JrzZVM_0:biae4am) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L173)):*
> ```EditorConfig
> 
> # CA1725: Parameter names should match base declaration
> dotnet_diagnostic.CA1725.severity = warning
> ```

This should be a suggestion.
___
*[`.editorconfig` line 198 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcgCHfAuzF8tL-VFBE:-Nqd1OUiCDBY3Byoyp3N:brgbej1) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L198)):*

This one should be a warning...
</blockquote></details>

I would leave this as a warning.
___
*[`src/Serval.ApiServer/Program.cs` line 3 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKHNd6s53_8Q8R445:-NqcKHNd6s53_8Q8R446:b-g96aaw) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.ApiServer/Program.cs#L3)):*
> ```Smalltalk
> namespace Serval.ApiServer;
> 
> #pragma warning disable CA1724 // Conflicts with SIL.Program
> ```

I really would prefer not to disable warnings. Is there no better way to fix this?
___
*[`src/Serval.ApiServer/Serval.ApiServer.csproj` line 11 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKYw29R0ZUgT8D_qr:-NqcKYw29R0ZUgT8D_qs:baappak) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.ApiServer/Serval.ApiServer.csproj#L11)):*
> ```XML
> 		<RunPostBuildEvent>OnBuildSuccess</RunPostBuildEvent>
> 		<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
> 		<GenerateDocumentationFile>true</GenerateDocumentationFile>
> ```

If we aren't using the documentation files, then we don't need to generate them. We won't need to suppress the related warnings as well.
___
*[`src/Serval.Shared/Services/ZipContainer.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd2okiEd_jnAJflYuZ:-Nqd2okiEd_jnAJflYu_:bwtcfcs) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.Shared/Services/ZipContainer.cs#L8)):*
> ```Smalltalk
> public class ZipContainer(string fileName) : DisposableBase, IZipContainer
> {
> #pragma warning disable CA2213 // Dispose in DisposeManagedResources
> ```

There's got to be a better way to handle this.
___
*[`src/SIL.DataAccess/MongoRepository.cs` line 11 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd5BFs1DSH-fdYqraz:-Nqd5BFs1DSH-fdYqrb-:b-hg9aql) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/MongoRepository.cs#L11)):*
> ```Smalltalk
>     public async Task<T?> GetAsync(Expression<Func<T, bool>> filter, CancellationToken cancellationToken = default)
>     {
>         if (_context.Session is not null)
> ```

Why were some braces removed for multiline `if` statements and some added?
___
*[`src/SIL.DataAccess/MongoSubscription.cs` line 109 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd5Vze3mwD9noAx_SO:-Nqd5Vze3mwD9noAx_SP:b-3f0htr) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/MongoSubscription.cs#L109)):*
> ```Smalltalk
>     protected override void DisposeManagedResources()
>     {
>         _context.Dispose();
> ```

This shouldn't be disposed here. These kinds of changes should be made in a separate PR.

johnml1135:
*[`.editorconfig` line 82 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcIUlcBb1ZUcXy0rXg:-NqdCo7S6wrdCSkwvcT5:b-qh39p2) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L82)):*

I don't like the `s_` for static fields. Static read-only fields are essentially constants, so I prefer them to use pascal case.
</blockquote></details>

https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions

-   Static fields start with `s_`. This convention isn't the default Visual Studio behavior, nor part of the [Framework design guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields), but is [configurable in editorconfig](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules).

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules

### Example: `const` modifier includes `static` and `readonly`
Consider the following two naming rules:
1.  Constant fields are PascalCase.
2.  Non-public `static` fields are s\_camelCase.
Rule 2 is more specific and takes precedence, so all non-public constant fields are s\_camelCase. To resolve the issue, you can define an intersection rule: non-public constant fields are PascalCase.

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members
❌ DO NOT use a prefix for field names.
For example, do not use "g\_" or "s\_" to indicate static fields.

johnml1135:
*[`.editorconfig` line 82 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcIUlcBb1ZUcXy0rXg:-NqdDOpIBfd0RD35wS4f:bi1lxbh) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L82)):*

https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions

-   Static fields start with `s_`. This convention isn't the default Visual Studio behavior, nor part of the [Framework design guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields), but is [configurable in editorconfig](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules).

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules

### Example: `const` modifier includes `static` and `readonly`
Consider the following two naming rules:
1.  Constant fields are PascalCase.
2.  Non-public `static` fields are s\_camelCase.
Rule 2 is more specific and takes precedence, so all non-public constant fields are s\_camelCase. To resolve the issue, you can define an intersection rule: non-public constant fields are PascalCase.

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members
❌ DO NOT use a prefix for field names.
For example, do not use "g\_" or "s\_" to indicate static fields.
</blockquote></details>

Ghaaa!  Microsoft contradicts itself!  What should we do?!?!?!?!
Hmm... the design principle of not adding [a-z]_ as a prefix appears sensible.  The only fixed prefixes and appended phrases are `I` for interfaces and `Async` for async calls.  Let's just use Pascal.

johnml1135:
*[`.editorconfig` line 173 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd4OMI7d23JrzZVM_-:-NqdGGIcAWZ1UMSoKyv_:bxeq6wg) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L173)):*

This should be a suggestion.
</blockquote></details>

Makes sense - tell the user but don'e mandate.  EngeineService.cs is a good example.

johnml1135:
*[`src/Serval.ApiServer/Program.cs` line 3 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKHNd6s53_8Q8R445:-NqdGpAB9tO9xL2M6EyO:bwwx84v) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.ApiServer/Program.cs#L3)):*

I really would prefer not to disable warnings. Is there no better way to fix this?
</blockquote></details>

I don't see a way: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1724.  Unless we want to remove the pattern (anti-pattern) of using globals, disabling the warning may be best.

johnml1135:
*[`src/Serval.ApiServer/Serval.ApiServer.csproj` line 11 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKYw29R0ZUgT8D_qr:-NqdK28SADh2NOyxZ65d:b-scq1i8) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.ApiServer/Serval.ApiServer.csproj#L11)):*

If we aren't using the documentation files, then we don't need to generate them. We won't need to suppress the related warnings as well.
</blockquote></details>

CSC : warning EnableGenerateDocumentationFile: Set MSBuild property 'GenerateDocumentationFile' to 'true' in project file to enable IDE0005 (Remove unnecessary usings/imports) on build (https://github.com/dotnet/roslyn/issues/41640) [/home/johnml1135/repos/serval/tests/Serval.ApiServer.IntegrationTests/Serval.ApiServer.IntegrationTests.csproj]

johnml1135:
*[`src/Serval.Shared/Services/ZipContainer.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd2okiEd_jnAJflYuZ:-NqdKXTX-6IbSU-X_Pn1:b-ijchvl) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.Shared/Services/ZipContainer.cs#L8)):*

There's got to be a better way to handle this.
</blockquote></details>

I think it is part of using the Disposable base.  When was that written?  Can we get rid of the DisposibleBase in Serval - doesn't dotnet have something standard that fills this need?

johnml1135:
*[`src/Serval.Shared/Services/ZipContainer.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd2okiEd_jnAJflYuZ:-NqdLDyv2cN7UUuV2GDz:b59jcv8) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.Shared/Services/ZipContainer.cs#L8)):*

I think it is part of using the Disposable base.  When was that written?  Can we get rid of the DisposibleBase in Serval - doesn't dotnet have something standard that fills this need?
</blockquote></details>

Will these updates work - remove DisposableBase, implement IDisposable and Dispose() and call GC.SuppressFinalize(this)?  If they work, I will use for all instances.

johnml1135:
*[`src/SIL.DataAccess/MongoRepository.cs` line 11 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd5BFs1DSH-fdYqraz:-NqdLk6iBP7-5WzN4PuF:b6v0le1) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/MongoRepository.cs#L11)):*

Why were some braces removed for multiline `if` statements and some added?
</blockquote></details>

I was wanting to remove most of them.  I can revert them.  Machine will not have these changes.

johnml1135:
*[`src/SIL.DataAccess/MongoSubscription.cs` line 109 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd5Vze3mwD9noAx_SO:-NqdLwls7iaci7hYUq0s:begcxyk) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/MongoSubscription.cs#L109)):*

This shouldn't be disposed here. These kinds of changes should be made in a separate PR.
</blockquote></details>

One option is to start this one from scratch as well - copy over the editorconfig and fix minimally.  What do you think?

johnml1135 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312)* status: 85 of 92 files reviewed, 10 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`.editorconfig` line 27 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKvTR8frYU13jAGB2:-NqdCMfW-i3zU255YHe-:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L27)):*

I think this is already specified elsewhere.
</blockquote></details>

Done.
___
*[`.editorconfig` line 198 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcgCHfAuzF8tL-VFBE:-NqdGTpi7ZvSJIfOObrX:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L198)):*

I would leave this as a warning.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.DataAccess/MongoSubscription.cs` line 109 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd5Vze3mwD9noAx_SO:-NqdPK44DnJ3Sp1kmBKw:b-sq666t) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/MongoSubscription.cs#L109)):*

One option is to start this one from scratch as well - copy over the editorconfig and fix minimally.  What do you think?
</blockquote></details>

I removed that dispose and checked - I did not add any more dispose in Serval - just a `GC.SuppressFinalize(this);` in ZipContainer.

ddaspit (commented):
> Reviewed 7 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312)* status: 91 of 92 files reviewed, 5 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`.editorconfig` line 82 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcIUlcBb1ZUcXy0rXg:-NqhLFFn8nnaNGgD5M1Z:bt95oiu) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L82)):*

Ghaaa!  Microsoft contradicts itself!  What should we do?!?!?!?!
Hmm... the design principle of not adding [a-z]_ as a prefix appears sensible.  The only fixed prefixes and appended phrases are `I` for interfaces and `Async` for async calls.  Let's just use Pascal.
</blockquote></details>

To tell you the truth, I never create static fields that aren't also read-only (essentially constants). To me that is a code smell.
___
*[`src/Serval.ApiServer/Program.cs` line 3 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKHNd6s53_8Q8R445:-NqhORLZ23uP7_u_qMcB:bk8s7hq) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.ApiServer/Program.cs#L3)):*

I don't see a way: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1724.  Unless we want to remove the pattern (anti-pattern) of using globals, disabling the warning may be best.
</blockquote></details>

Can you just remove the global using that is causing this conflict?
___
*[`src/Serval.ApiServer/Serval.ApiServer.csproj` line 11 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKYw29R0ZUgT8D_qr:-NqhMu22-7I2K3-grNxO:b-wpdy2) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.ApiServer/Serval.ApiServer.csproj#L11)):*

CSC : warning EnableGenerateDocumentationFile: Set MSBuild property 'GenerateDocumentationFile' to 'true' in project file to enable IDE0005 (Remove unnecessary usings/imports) on build (https://github.com/dotnet/roslyn/issues/41640) [/home/johnml1135/repos/serval/tests/Serval.ApiServer.IntegrationTests/Serval.ApiServer.IntegrationTests.csproj]
</blockquote></details>

That is interesting. I wonder why. In any case, that's fine.
___
*[`src/Serval.Shared/Services/ZipContainer.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd2okiEd_jnAJflYuZ:-NqhNSy70rxh3ZR6XJtG:bka34ij) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.Shared/Services/ZipContainer.cs#L8)):*

Will these updates work - remove DisposableBase, implement IDisposable and Dispose() and call GC.SuppressFinalize(this)?  If they work, I will use for all instances.
</blockquote></details>

I think that's fine. It is better than adding ugly suppress warning pragmas in every disposable class. It is a bit annoying that we can't use the `DisposableBase` class that implements the dispose pattern for us. Maybe it would be better to just turn the check into a suggestion.
___
*[`src/SIL.DataAccess/MongoRepository.cs` line 11 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd5BFs1DSH-fdYqraz:-NqhO0SM1iLODevN4aIb:bp4rwdj) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/MongoRepository.cs#L11)):*

I was wanting to remove most of them.  I can revert them.  Machine will not have these changes.
</blockquote></details>

Yes, please revert them. We only want to remove the braces on `if...else` statements where every statement body is only a single line.

johnml1135:
*[`src/Serval.Shared/Services/ZipContainer.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd2okiEd_jnAJflYuZ:-NqiCwAe169BF6Dj_-Gs:b-migk05) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.Shared/Services/ZipContainer.cs#L8)):*

I think that's fine. It is better than adding ugly suppress warning pragmas in every disposable class. It is a bit annoying that we can't use the `DisposableBase` class that implements the dispose pattern for us. Maybe it would be better to just turn the check into a suggestion.
</blockquote></details>

What is the difference between DisposableBase and IDisposable?  Why was DisposableBase created in the first place?

johnml1135:
*[`.editorconfig` line 82 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcIUlcBb1ZUcXy0rXg:-NqiGBva1-LTVUugu3vZ:b-ec5je3) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/.editorconfig#L82)):*

To tell you the truth, I never create static fields that aren't also read-only (essentially constants). To me that is a code smell.
</blockquote></details>

All to Pascal.  Also, we are checking for CA2211.

johnml1135:
*[`src/SIL.DataAccess/MongoRepository.cs` line 11 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd5BFs1DSH-fdYqraz:-NqiN-420iuMNBHVUCIz:b-bw8e14) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/MongoRepository.cs#L11)):*

Yes, please revert them. We only want to remove the braces on `if...else` statements where every statement body is only a single line.
</blockquote></details>

Updated.

johnml1135:
*[`src/Serval.ApiServer/Program.cs` line 3 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKHNd6s53_8Q8R445:-NqiR9eA7sN5y4ns1kFn:b6im0bo) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.ApiServer/Program.cs#L3)):*

Can you just remove the global using that is causing this conflict?
</blockquote></details>

Hmmm.  If we get rid of DisposableBase, we may be able to get rid of SIL.Core which has the SIL.Program...

ddaspit (commented):
> Reviewed 14 of 14 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312)* status: 91 of 92 files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval.ApiServer/Program.cs` line 3 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKHNd6s53_8Q8R445:-NqmVH9T3WFTWCBIwR7h:byabz7q) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.ApiServer/Program.cs#L3)):*

Hmmm.  If we get rid of DisposableBase, we may be able to get rid of SIL.Core which has the SIL.Program...
</blockquote></details>

I think we should just make this rule a suggestion.
___
*[`src/Serval.Shared/Services/ZipContainer.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd2okiEd_jnAJflYuZ:-NqmTVnJAz2U8a3KinPz:befub8g) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.Shared/Services/ZipContainer.cs#L8)):*

What is the difference between DisposableBase and IDisposable?  Why was DisposableBase created in the first place?
</blockquote></details>

`DisposableBase` implements the [dispose pattern](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) properly. It can be a bit tricky to get right and it makes it very clear where things should be disposed by providing the `DisposeManagedResources` and `DisposeUnmanagedResources` methods.

johnml1135:
*[`src/Serval.Shared/Services/ZipContainer.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd2okiEd_jnAJflYuZ:-NqmgbsQ9fKBl2vVmtH5:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.Shared/Services/ZipContainer.cs#L8)):*

`DisposableBase` implements the [dispose pattern](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) properly. It can be a bit tricky to get right and it makes it very clear where things should be disposed by providing the `DisposeManagedResources` and `DisposeUnmanagedResources` methods.
</blockquote></details>

Done.

johnml1135:
*[`src/Serval.ApiServer/Program.cs` line 3 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcKHNd6s53_8Q8R445:-Nqmgf3m0uMPznD3wybY:blggctd) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.ApiServer/Program.cs#L3)):*

I think we should just make this rule a suggestion.
</blockquote></details>

I'll make it a suggestion and revert changes.

johnml1135:
*[`src/Serval.Shared/Services/ZipContainer.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-Nqd2okiEd_jnAJflYuZ:-Nqmgqrm0DAxxnKoDXYK:b-a1xtl6) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/Serval.Shared/Services/ZipContainer.cs#L8)):*

Done.
</blockquote></details>

I'll revert all to DisposableBase.

Enkidu93 (approved):
> Reviewed 14 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.DataAccess/ExpressionHelper.cs` line 32 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcgYknEIPgkBrDnv38:-Nqmrg6S5E6A-7_e3pf_:b-gxis8o) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/ExpressionHelper.cs#L32)):*

Renamed to TConstant.  It was complaining about https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1715.
</blockquote></details>

I see - that makes sense.

johnml1135 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`src/SIL.DataAccess/ExpressionHelper.cs` line 32 at r2](https://reviewable.io/reviews/sillsdev/serval/312#-NqcgYknEIPgkBrDnv38:-Nqn14sj3hdVWb7kTisR:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/f923e7338a4c770c67d55bb01aff6ad61c79d563/src/SIL.DataAccess/ExpressionHelper.cs#L32)):*

I see - that makes sense.
</blockquote></details>

Done.

ddaspit (commented):
One last thing, please add `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` to all csproj files.

> Reviewed 14 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312#-:-NqnGbs4Cc39pQ5KMU7Y:bc0urq9)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

johnml1135 (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312#-:-NqnI08T2WaD67XH95P4:bb6un1j)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

ddaspit (approved):
:lgtm:

> Reviewed 15 of 15 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/312#-:-NqnIOlb-PfSqq8r3tH3:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

Issue #361, opened 2024-04-15
Optimize Garbage Collector settings


johnml1135:
The GC settings need to be aligned to the alert thresholds.  Since we alert at 90%, we can set the threshold to 80%.  This should queue is into whether there actually is an issue or not (long term with memory).  This is a spin off of #338.

johnml1135:
We can just live with the OOM's right now.  If we need to, we can increase RAM.

PR #363, opened 2024-04-15
Update garbage collection settings

johnml1135:
https://github.com/sillsdev/serval/issues/361

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/363)

### DIFF
--- a/deploy/serval/templates/prometheus-rules.yaml
+++ b/deploy/serval/templates/prometheus-rules.yaml
@@ -38,12 +38,12 @@ spec:
         - alert: memory-near-limit
           annotations:
             description: >-
-              '{{ "{{ $labels.container }}" }} is over 80% memory.'
+              '{{ "{{ $labels.container }}" }} is over 95% memory.'
           expr: >-
             min (container_memory_working_set_bytes{image!="", namespace="{{ .Values.namespace }}", container!="POD" })
             by (container, namespace) / on (container, namespace)
             min (kube_pod_container_resource_limits{resource="memory", namespace="{{ .Values.namespace }}", container!="POD"})
-            by (container, namespace) * 100 >= 80
+            by (container, namespace) * 100 >= 95
           for: 0s
           labels:
             severity: warning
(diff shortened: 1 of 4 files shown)


ddaspit (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/363)* status: all files reviewed, 1 unresolved discussion (waiting on @johnml1135)

___
*[`src/Serval.ApiServer/runtimeconfig.template.json` line 2 at r1](https://reviewable.io/reviews/sillsdev/serval/363#-NvYaWMR6Lckfw9OGn6N:-NvYaWMR6Lckfw9OGn6O:bhqioa2) ([raw file](https://github.com/sillsdev/serval/blob/775d5f14e8b53fa0894810cfdae2cb2ccdb5717b/src/Serval.ApiServer/runtimeconfig.template.json#L2)):*
> ```JSON
> {
>     "configProperties": {
> ```

I'm not sure I understand what issue we are trying to solve with this change. Also, we should not mess with the default GC settings. It will probably result in a negative impact on performance.

johnml1135:
*[`src/Serval.ApiServer/runtimeconfig.template.json` line 2 at r1](https://reviewable.io/reviews/sillsdev/serval/363#-NvYaWMR6Lckfw9OGn6N:-NvaluO_6aRXxQa3d-3O:bitiz7b) ([raw file](https://github.com/sillsdev/serval/blob/775d5f14e8b53fa0894810cfdae2cb2ccdb5717b/src/Serval.ApiServer/runtimeconfig.template.json#L2)):*

I'm not sure I understand what issue we are trying to solve with this change. Also, we should not mess with the default GC settings. It will probably result in a negative impact on performance.
</blockquote></details>

Yes, but we are only using 2-3% (normally) of the CPU but sometimes crash because of running out of memory.  The only essential one of the lot is  `"System.GC.HighMemoryPercent": 80`.  If you are concerned about the decrease in performance, I would be ok just setting that one and seeing how it performs.

johnml1135:
We are holding of for this one right now...

PR #394, opened 2024-05-23
Treat empty ScriptureRange like null

### LINKED ISSUE #386: Exception thrown when no training data is selected
Source was Spanish, and target was English, so no training data should be required in order to generate a draft. However, the build failed with this:

> Message: The build configuration was invalid. Status: 400 Response: "The scripture range is not valid: Invalid syntax. If you are providing multiple selections, e.g. a range of books followed by a selection of chapters from a book, separate each selection with a semicolon."
Translation Engine Id: 66356177d874d81aa4ebaa0e

Enkidu93:
Fixes #386

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/394)

### DIFF
--- a/deploy/qa-ext-values.yaml
+++ b/deploy/qa-ext-values.yaml
@@ -1,6 +1,6 @@
 externalHost: qa.serval-api.org
 environment: Production
-deploymentVersion: '1.4.QA3'
+deploymentVersion: '1.4.QA4'
 alertEmail: ext-qa-serval-alerts@languagetechnology.org
 emailsToAlert: john_lambert@sil.org
 enableTls: true
@@ -8,8 +8,8 @@ namespace: serval
 auth0Domain: dev-sillsdev.auth0.com
 lokiTenent: serval-tenant
 lokiUrl: http://loki-distributed-gateway.loki.svc.cluster.local
-servalImage: ghcr.io/sillsdev/serval:1.4.3
-machineImage: ghcr.io/sillsdev/machine:3.9.3
+servalImage: ghcr.io/sillsdev/serval:1.4.4
+machineImage: ghcr.io/sillsdev/machine:3.9.4
 ClearMLDockerImage: ghcr.io/sillsdev/machine.py:1.4.0
 ClearMLQueue: production
 MongoConnectionPrefix: qa_
--- a/src/Serval.Translation/Services/EngineService.cs
+++ b/src/Serval.Translation/Services/EngineService.cs
@@ -234,7 +234,7 @@ public class EngineService(
                                 pretranslateCorpus.TextIds is null && pretranslateCorpus.ScriptureRange is null;
                             if (pretranslateCorpus.TextIds is not null)
                                 corpus.PretranslateTextIds.Add(pretranslateCorpus.TextIds);
-                            if (pretranslateCorpus.ScriptureRange is not null)
+                            if (!string.IsNullOrEmpty(pretranslateCorpus.ScriptureRange))
                             {
                                 if (
                                     c.TargetFiles.Count > 1
@@ -264,7 +264,7 @@ public class EngineService(
                             corpus.TrainOnAll = trainingCorpus.TextIds is null && trainingCorpus.ScriptureRange is null;
                             if (trainingCorpus.TextIds is not null)
                                 corpus.TrainOnTextIds.Add(trainingCorpus.TextIds);
-                            if (trainingCorpus.ScriptureRange is not null)
+                            if (!string.IsNullOrEmpty(trainingCorpus.ScriptureRange))
                             {
                                 if (
                                     c.TargetFiles.Count > 1
--- a/tests/Serval.Translation.Tests/Services/EngineServiceTests.cs
+++ b/tests/Serval.Translation.Tests/Services/EngineServiceTests.cs
@@ -413,6 +413,7 @@ public class EngineServiceTests
                             Id = "corpus1",
                             SourceLanguage = "es",
                             TargetLanguage = "en",
+                            TrainOnAll = false,
                             SourceFiles =
                             {
                                 new V1.CorpusFile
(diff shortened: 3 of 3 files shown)


ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/394)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/Serval.Translation/Services/EngineService.cs` line 237 at r1](https://reviewable.io/reviews/sillsdev/serval/394#-NybR43d-R78KYbZxMTL:-NybR43d-R78KYbZxMTM:bnufakg) ([raw file](https://github.com/sillsdev/serval/blob/637e83b6f8cfbad9edaf783463d5b16af131c19a/src/Serval.Translation/Services/EngineService.cs#L237)):*
> ```Smalltalk
>                             if (pretranslateCorpus.TextIds is not null)
>                                 corpus.PretranslateTextIds.Add(pretranslateCorpus.TextIds);
>                             if (!string.IsNullOrEmpty(pretranslateCorpus.ScriptureRange))
> ```

Does the `ScriptureRangeParser` crash when you pass in an empty string? I feel like an empty string should mean don't include any books in the same way that an empty `textids` list means don't include any books.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/394)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`src/Serval.Translation/Services/EngineService.cs` line 237 at r1](https://reviewable.io/reviews/sillsdev/serval/394#-NybR43d-R78KYbZxMTL:-NyebsilDdtl0WlqGX0Q:b23y2z1) ([raw file](https://github.com/sillsdev/serval/blob/637e83b6f8cfbad9edaf783463d5b16af131c19a/src/Serval.Translation/Services/EngineService.cs#L237)):*

Does the `ScriptureRangeParser` crash when you pass in an empty string? I feel like an empty string should mean don't include any books in the same way that an empty `textids` list means don't include any books.
</blockquote></details>

Yes, it raises an exception. I went back and forth as to what made more sense. `""` doesn't properly seem like a valid scripture range. I can make the change in the parser if you'd prefer. This is just a simple way to make sure that `null`/`""` have the same exact behavior from this point in the pipeline on. Also, if I make that change, I'll need to also make the matching change in machine.py. Let me know.

ddaspit (commented):
You should add a unit test to cover this case.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/394#-:-Nyey9KB2ZhSIbLGNTZX:bglr8hf)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

___
*[`src/Serval.Translation/Services/EngineService.cs` line 237 at r1](https://reviewable.io/reviews/sillsdev/serval/394#-NybR43d-R78KYbZxMTL:-Nyex17d3J3Iv5zrj_k3:bob2jzv) ([raw file](https://github.com/sillsdev/serval/blob/637e83b6f8cfbad9edaf783463d5b16af131c19a/src/Serval.Translation/Services/EngineService.cs#L237)):*

Yes, it raises an exception. I went back and forth as to what made more sense. `""` doesn't properly seem like a valid scripture range. I can make the change in the parser if you'd prefer. This is just a simple way to make sure that `null`/`""` have the same exact behavior from this point in the pipeline on. Also, if I make that change, I'll need to also make the matching change in machine.py. Let me know.
</blockquote></details>

`ScripureRangeParser` should be able to handle an empty string, but we don't need to fix that right now. This is sufficient to fix the bug.

Enkidu93 (commented):
I apologize: Somehow, I missed your comment about the unit test. Done.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/394#-:-Nz-GXBZ67Qnpdf-pUzt:bx1k6w4)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @ddaspit)

ddaspit (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/394)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`tests/Serval.Translation.Tests/Services/EngineServiceTests.cs` line 411 at r2](https://reviewable.io/reviews/sillsdev/serval/394#-Nz40a8oAZeXeQjYkm5j:-Nz40a8oAZeXeQjYkm5k:b-qrd0iq) ([raw file](https://github.com/sillsdev/serval/blob/f1cc552ebc3c14073e45716b2776164fc7bbec60/tests/Serval.Translation.Tests/Services/EngineServiceTests.cs#L411)):*
> ```Smalltalk
>                     Corpora =
>                     {
>                         new V1.Corpus
> ```

One of the key things is that `TrainOnAll` is `false`. Please explicitly set that property so it is clear.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/394)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`tests/Serval.Translation.Tests/Services/EngineServiceTests.cs` line 411 at r2](https://reviewable.io/reviews/sillsdev/serval/394#-Nz40a8oAZeXeQjYkm5j:-Nz4nMb852EuaSAaK3Ir:bb6un1j) ([raw file](https://github.com/sillsdev/serval/blob/f1cc552ebc3c14073e45716b2776164fc7bbec60/tests/Serval.Translation.Tests/Services/EngineServiceTests.cs#L411)):*

One of the key things is that `TrainOnAll` is `false`. Please explicitly set that property so it is clear.
</blockquote></details>

Done

ddaspit (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/394)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

PR #419, opened 2024-06-26
Pretranslate into source usfm

johnml1135:
Ready to roll.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/419)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -26,6 +26,7 @@
     "Protobuf",
     "ptcc",
     "Rebinder",
+    "Stylesheet",
     "upserted",
     "USFM"
   ],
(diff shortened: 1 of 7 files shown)


ddaspit (changes requested):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/419)* status: all files reviewed, 1 unresolved discussion (waiting on @johnml1135)

___
*[`src/Serval/src/Serval.Translation/Services/PretranslationService.cs` line 125 at r1](https://reviewable.io/reviews/sillsdev/serval/419#-O0jQ2NKB40hI94UjnrK:-O0jQ2NKB40hI94UjnrL:b-gom8wb) ([raw file](https://github.com/sillsdev/serval/blob/55f139581e5872e4b337630c9e822da4d61c8504/src/Serval/src/Serval.Translation/Services/PretranslationService.cs#L125)):*
> ```Smalltalk
>         if (sourceUsfm is not null)
>         {
>             IReadOnlyList<(IReadOnlyList<ScriptureRef>, string)> targetTextAsPretranslations = [];
> ```

I don't think that all of this is necessary. The pretranslations are created from the source text, so they should match the source structure exactly. SF doesn't want to copy the existing target text over, so the existing code should be sufficient to fulfill the requirement.

PR #420, opened 2024-07-03
Add "UseSourceUsfm" option to USFM endpoint.

### LINKED ISSUE #405: Choice to always use source USFM for generating new USFM
Currently the choice of whether to use source or target USFM as the template for inserting in the pretranslations is not given to the user.  This issue is to allow that choice to be made at the Serval level - this is a spin off from https://github.com/sillsdev/serval/issues/398 from @pmachapman's comments. 

Issues:
* I assume that the existing target verses may be able to be placed in the source USFM template - or will we only allow inserting pretranslations?
* Will this be a 5th option for "pretranslations only", etc., such as "pretranslations in source USFM", or will it be a separate field that can have the 4 options?
* Is this really needed or would better USFM handling and error visibility be the main issue?

johnml1135:
fixes #405.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/420)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -26,6 +26,7 @@
     "Protobuf",
     "ptcc",
     "Rebinder",
+    "stylesheet",
     "upserted",
     "USFM"
   ],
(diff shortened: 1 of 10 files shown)


johnml1135 (commented):
> Reviewed 1 of 7 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/420)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

PR #437, opened 2024-07-17
Remove timeout on MessageOutboxDeliveryService

### LINKED ISSUE #422: Better database indexing for pretranslations
Based upon https://www.mongodb.com/docs/manual/tutorial/create-indexes-to-support-queries/#std-label-create-indexes-to-support-queries, we should use multi-indexing for pretranslations.  Specifically, it appears that we should use a multi-index of the engineId and the Scripture reference to be able to get the verses without having to scan all pretranslations.

Enkidu93:
Fixes #422 

I've gone ahead and added the recommended index on pretranslations which should speed up queries there significantly, but ultimately, it wasn't the queries on pretranslations that were taxing mongo and causing it to send alerts. 

Rather, it has to do with a bug in the `MessageOutboxDeliveryService`'s subscription. A timeout of 10 seconds is used when waiting for a new message to appear. However, when looping back to wait again, it's looking for changes from the same time onward since the timestamp which it uses as a placeholder only updates when there's been a change (i.e., not when it's timed out). For now, I've simply removed the timeout which fits with usage elsewhere. Now it should wait until there's been a change without rerunning the changeStream command. In recent versions of Mongo, it's also possible to resume changeStreams. It may be better (if we have situations in which we want a timeout but will also re-wait for changes) to use the resumeToken strategy to resume rather than restart.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/437)

### DIFF
--- a/docker-compose.withatlas.yml
+++ b/docker-compose.withatlas.yml
@@ -14,8 +14,8 @@ services:
       - ASPNETCORE_Kestrel__Endpoints__Http__Url=http://*:80
       - ASPNETCORE_Kestrel__Endpoints__Http2__Url=http://*:81
       - ASPNETCORE_Kestrel__Endpoints__Http2__Protocols=Http2
-      - ASPNETCORE_ConnectionStrings__Hangfire=${MONGO_CONNECTION_STRING:?connection string needed}/${MONGO_CONNECTION_PREFIX:?mongo prefix needed}serval_jobs
-      - ASPNETCORE_ConnectionStrings__Mongo=${MONGO_CONNECTION_STRING}/${MONGO_CONNECTION_PREFIX}serval
+      - ASPNETCORE_ConnectionStrings__Hangfire=${MONGO_CONNECTION_STRING:?connection string needed}/${MONGO_CONNECTION_PREFIX:?mongo prefix needed}serval_jobs${MONGO_EXTENSIONS}
+      - ASPNETCORE_ConnectionStrings__Mongo=${MONGO_CONNECTION_STRING}/${MONGO_CONNECTION_PREFIX}serval${MONGO_EXTENSIONS}
       - ASPNETCORE_Translation__Engines__0__Type=Echo
       - ASPNETCORE_Translation__Engines__0__Address=http://echo
       - ASPNETCORE_Translation__Engines__1__Type=SmtTransfer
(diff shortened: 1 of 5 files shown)


johnml1135:
@Enkidu93, do you have a reference to any documentation that uses the pattern that you are recommending?

johnml1135:
That is, do you have documentation that this solution will work, or did you test it out empirically?

johnml1135:
Also, can you change the dot net code to correct the pre-translation indexing as well?

Enkidu93 (commented):
Here's a section where the documentation discusses techniques for resuming/starting change streams: https://www.mongodb.com/docs/manual/changeStreams/#resumeafter-for-change-streams

In the same area, you can see information about the method we're using which is to start from a particular time. If you look at the logs on Atlas, you can see that the change stream command is running from the same start time every time until an insertion has happened. And the command reruns because of the timeout. Inasmuch as I can verify locally, I have. Of course, it's not quite as convenient to see into things on the local instance as it is via Atlas. 

Note: Since the outbox code is only on QA for the time being, the issue is only visible there. I'm concerned about pushing current QA code to production given the strain this puts on Mongo, but at the same time, this issue should actually be less problematic the more activity there is in the outbox. 

In regards to updating the .NET in some way to do with indexing, I'm not sure what you mean? I created the index in the db. What do I need to do in the code?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/437#-:-O22bIYq2I6P12Z1kf8F:b-k89927)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @ddaspit and @johnml1135)

johnml1135:
The removal of the timeout appears like the right choice - I didn't understand the rational behind the timeout (it was only used in Machine for the Locks to not grind the whole software to a halt).  

As for indexing, I think we would want a [compond index](https://www.mongodb.com/docs/drivers/csharp/current/fundamentals/indexes/#compound-indexes).  That way we can sort and index by (at least) the EngineId and CorpusId, if not also the Scripture reference.  Here is [how to implement it in C#](https://stackoverflow.com/questions/14904733/creating-a-compound-index-in-c-sharp).  

I don't know the best way to add it - would we be able to add a new parameter of type `CreateIndexModel<T>` to `AddRepository` that can be applied to the collection as it is created.

johnml1135 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/437)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

johnml1135:
Check that - CreateIndexModels are already being created for the pretranslations - but they might not be ideal.  I pushed some code - what do you think?

Enkidu93 (commented):
OK!

Yes, I already created a compound index in Atlas on pretranslations on the engineRef and modelRevision (the recommended index given the queries from SF although it looks like maybe a compound index that includes the textId as well could help so I'll go ahead and make that as well). But I see what you're saying: You are  wanting me to verify that those indices exist programmatically in the the mongo instance on startup - I can do that. For Atlas, it probably isn't as necessary, but it won't hurt.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/437#-:-O25-1Q1-YU--ckT61Kj:b-s6wlas)* status: 1 of 4 files reviewed, all discussions resolved (waiting on @ddaspit)

johnml1135:
Can you review the code that I pushed and see if it resolves the issues?  I also checked for other places in the code that look like they should have a multi-index but don't.

Enkidu93:
I think this is the arrangement we want (although I'd like to spend some more time peeking at all the indices at some point and giving it more thought). We don't want to remove the simple indices even if we do add compound indices unless the simple index is a prefix of the compound index. We also want to avoid unnecessary indices because they take up memory and make writes more expensive. I went ahead and looked at how frequently the indices are being used and removed ones that have never been used. I removed redundant simple indices and added the suggested compound index on pretranslations.

johnml1135:
Please test on docker-compose with Atlas.

Enkidu93 (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/437#-:-O273IRaDlnYPWWeKEUS:bg7igkv)* status: 1 of 5 files reviewed, all discussions resolved (waiting on @ddaspit)

johnml1135:
:lgtm:

johnml1135 (approved):
> Reviewed 3 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/437)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

Issue #442, opened 2024-07-25
Last verse USFM blank when source used for template 


johnml1135:
This is a spin off of #408 - it for some reason is still showing up when the source USFM is chosen to be the template.  Why?

pmachapman:
This bug is only fixed on 1.5.QA5 when the template is set to Target. If the template is set to Source, the bug still occurs.

On 1.5.QA5, this endpoint https://qa.serval-api.org/api/v1/translation/engines/662e83d5d3673074e658b6db/corpora/662e83dfd3673074e658b6de/pretranslations/SUS/usfm?text-origin=OnlyPretranslated&template=Target returns:
```
...
\v 64 And Daniel became great in the sight of the people from that day and forward.
\v 65 And king Astyages was laid to rest with his fathers, and Cyrus the Persian received his kingdom.
```

While https://qa.serval-api.org/api/v1/translation/engines/662e83d5d3673074e658b6db/corpora/662e83dfd3673074e658b6de/pretranslations/SUS/usfm?text-origin=OnlyPretranslated&template=Source returns:
```
...
\v 64 And Daniel became great in the sight of the people from that day and forward.
\v 65
```

Enkidu93:
I think I've figured the issue out. I'll try and get a PR in Monday.

Enkidu93:
Alright, so the issue is versification-related like I thought: This seems to be connected to the fact that Susanna properly only has 64 verses. But in the Vulgate versification (used in the source and target above), Daniel 13:1-64 = SUS 1:1-64, but Daniel 13:65 = BEL 1:1. In other words, there shouldn't be 65 verses for Susanna. I suspect the reason that it's working with `template=Target` is because it's using relaxed `ScriptureRef`s. Why are we only using relaxed for `template=Target` and not `Source`, @ddaspit @johnml1135? - I think I was away for that PR. And how should we handle this exactly?

ddaspit:
@Enkidu93 It doesn't have to do with relaxed `ScriptureRef`s. It has to do with switching between the source and target versifications. Can you send me the source and target projects? I would like to see how the versifications are set up for each project.

Enkidu93:
> @Enkidu93 It doesn't have to do with relaxed `ScriptureRef`s. It has to do with switching between the source and target versifications. Can you send me the source and target projects? I would like to see how the versifications are set up for each project.

OK, that was actually my first guess, but the versifications are the same, so shouldn't it not matter?

Sure.
[...]

PR #471, opened 2024-08-29
Prestranslation versification mismatch

Enkidu93:
Partial fix for https://github.com/sillsdev/serval/issues/442

This solution has been verified by a manual test. However, I'm not convinced that this is the best course of action: My first go at a fix was to parse the pretranslation refs as source refs (which is what they should be, right, given that the `Refs` property yields the source refs as long as there are source refs which should be the case for any verse that's being pretranslated?) and then change the versification to the target when using the target as a template, but that did not work (it just flipped the issue - i.e., fail when template is set to target, works when set to source). I believe that this is because `ScriptureRef.Parse()` is happy to construct a verse that doesn't actually exist within the project's versification. That is to say, it doesn't care that the project versification is incorrect. However, the `ChangeVersification` method does care. The solution I'm giving here does solve this issue, but I'm concerned that it will ultimately cause issues elsewhere (even though tests are passing) since if you set template to target, you'll be potentially parsing source versifications as target versifications. 

It seems to me that the best solution would be to parse the pretranslation refs as source refs and convert to target versification when using target as a template and let this be an outstanding issue since this is really the fault of the project. Let me know what you all think. I can't think of a good way to implement our own versification conversion that would account for this issue since we're in a situation in which the project versification is incorrect.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/471)

### DIFF
--- a/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs
@@ -400,7 +400,7 @@ public class PreprocessBuildJob : HangfireBuildJob<IReadOnlyList<Corpus>>
         {
             if (!row.IsTargetRangeStart && row.IsTargetInRange)
             {
-                refs.AddRange(row.Refs);
+                refs.AddRange(row.TargetRefs);
                 if (row.SourceText.Length > 0)
                 {
                     if (srcSegBuffer.Length > 0)
@@ -422,7 +422,7 @@ public class PreprocessBuildJob : HangfireBuildJob<IReadOnlyList<Corpus>>
                 }
 
                 textId = row.TextId;
-                refs.AddRange(row.Refs);
+                refs.AddRange(row.TargetRefs);
                 srcSegBuffer.Append(row.SourceText);
                 trgSegBuffer.Append(row.TargetText);
                 rowCount++;
--- a/src/Serval/src/Serval.Translation/Services/PretranslationService.cs
+++ b/src/Serval/src/Serval.Translation/Services/PretranslationService.cs
@@ -56,25 +56,26 @@ public class PretranslationService(
             targetFile.Filename
         );
 
-        IEnumerable<(IReadOnlyList<string> Refs, string Translation)> rawPretranslations = (
+        IEnumerable<(IReadOnlyList<ScriptureRef> Refs, string Translation)> pretranslations = (
             await GetAllAsync(engineId, modelRevision, corpusId, textId, cancellationToken)
-        ).Select(p => (p.Refs, p.Translation));
+        )
+            .Select(p =>
+                (
+                    Refs: (IReadOnlyList<ScriptureRef>)
+                        p.Refs.Select(r => ScriptureRef.Parse(r, targetSettings.Versification)).ToArray(),
+                    p.Translation
+                )
+            )
+            .OrderBy(p => p.Refs[0]);
 
         // Update the target book if it exists
         if (template is PretranslationUsfmTemplate.Auto or PretranslationUsfmTemplate.Target)
         {
             // the pretranslations are generated from the source book and inserted into the target book
             // use relaxed references since the USFM structure may not be the same
-            IEnumerable<(IReadOnlyList<ScriptureRef> Refs, string Translation)> pretranslations = rawPretranslations
-                .Select(p =>
-                    (
-                        (IReadOnlyList<ScriptureRef>)
-                            p.Refs.Select(r => ScriptureRef.Parse(r, targetSettings.Versification).ToRelaxed())
-                                .ToArray(),
-                        p.Translation
-                    )
-                )
-                .OrderBy(p => p.Item1[0]);
+            pretranslations = pretranslations.Select(p =>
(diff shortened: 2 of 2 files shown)


Enkidu93:
As per our chat in the stand up, I've made the changes we discussed to the code locally and am testing it now manually.

Enkidu93:
@ddaspit, is this what you had in mind? Just to verify (unless I've made a mistake), this would leave that 65th verse blank for both template as source and as target (judging from manual testing). Is that what you expected?

ddaspit (commented):
This looks correct, but it seems to work on my test. Check out my [test](https://github.com/sillsdev/machine/commit/d8be7032113f016f072eb7f1246ef00c9cacc9a0) and see if we are testing the same thing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/471#-:-O5_5J1p7vMlopbn9X8y:b-ub2rs5)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @johnml1135)

Enkidu93:
> This looks correct, but it seems to work on my test. Check out my [test](https://github.com/sillsdev/machine/commit/d8be7032113f016f072eb7f1246ef00c9cacc9a0) and see if we are testing the same thing.
> 
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/serval/471#-:-O5_5J1p7vMlopbn9X8y:b-ub2rs5)_ status: 0 of 2 files reviewed, all discussions resolved (waiting on @johnml1135)

One difference seems to be that in Serval, the ProjectTextUpdater is using the target project. If you use the target and the corresponding parameters to UpdateUsfm, it still fails. I still haven't figured out what the difference is when it's set to source.

ddaspit (approved):
From my testing, it works with `PretranslationUsfmTemplate.Source`, but not with `PretranslationUsfmTemplate.Target`, because of the error in the versification. If you are seeing the same thing, then it is working as expected.

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/471#-:-O5tPvje6xTOzVdOjEkk:b-1pc41x)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Enkidu93:
> From my testing, it works with `PretranslationUsfmTemplate.Source`, but not with `PretranslationUsfmTemplate.Target`, because of the error in the versification. If you are seeing the same thing, then it is working as expected.
> 
> > Reviewed 2 of 2 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/serval/471#-:-O5tPvje6xTOzVdOjEkk:b-1pc41x)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

OK, that's good to know. I reran an E2E manual test just to confirm, and no, it is not working for `PretranslationUsfmTemplate.Source` either. Still investigating. I'm sure I must be missing something silly.

ddaspit (commented):
This might be a silly question, but are you re-running the build for this E2E test or are you just retrieving the pretranslations from an already completed build? The bug fix impacts preprocessing, so you will need to run a full build, so that the pretranslations have the correct verse ref.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/471#-:-O5thKI03rl0t1STRfK5:b-4wrgyy)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Enkidu93:
> This might be a silly question, but are you re-running the build for this E2E test or are you just retrieving the pretranslations from an already completed build? The bug fix impacts preprocessing, so you will need to run a full build, so that the pretranslations have the correct verse ref.
> 
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/serval/471#-:-O5thKI03rl0t1STRfK5:b-4wrgyy)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

No, no silly questions at this point haha. But yes, I am rerunning the whole E2E. It's definitely something on the preprocess side. I keep getting side-tracked from debugging though.

Enkidu93:
OK! I've figured it out - kind of. I've edited your test, Damien, and it now fails. When using text ids, it works, but when using chapters, it does not. That's why I couldn't figure it out for the life of me! What's happening is that  ref`SUS 1:65` in the source is being mapped to ref `DAG 13:3` in the target. When it runs `IsInPretranslate`, the text id is still `SUS` regardless so if you just provide text ids, it works properly. But if you use chapters, it looks at `DAG`. This seems to me to be an error in the `AlignRows` function itself - it really shouldn't be aligning those verses under any circumstances (maybe `SUS 1:65` -> `DAG 13:65`, but not `13:3`). Now, it might just be that the versification being incorrect is causing this, but I do think this merits further investigation. What do you think is the best course of action, @ddaspit? We could put this PR through and open an issue to investigate and maybe add some more testing in general that involves cross-versification alignment and the deuterocanon - the disparity in verse references between `SUS 1:65` and `DAG 13:3` (when, if anything, it should be `65`) is concerning to me.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/471)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

PR #491, opened 2024-09-13
Avoid constantly attempting to acquire a lock

ddaspit:
- ignore expiration timeout on WaitForChangesAsync when there are queued writer locks

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/491)

### DIFF
--- a/src/Machine/src/Serval.Machine.Shared/Services/DistributedReaderWriterLock.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/DistributedReaderWriterLock.cs
@@ -73,8 +73,8 @@ public class DistributedReaderWriterLock(
                 RWLock? rwLock = sub.Change.Entity;
                 if (rwLock is not null && !rwLock.IsAvailableForReading())
                 {
-                    TimeSpan? timeout = default;
-                    if (rwLock.WriterLock?.ExpiresAt is not null)
+                    TimeSpan? timeout = null;
+                    if (rwLock.WriterQueue.Count == 0 && rwLock.WriterLock?.ExpiresAt is not null)
                     {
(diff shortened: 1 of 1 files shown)


johnml1135:
This appears to prevent the infinite cycling.  Still, if there is an expired writer on the queue (with a different lock Id), it will never be cleared away.  Such could happen if the database is inaccessible when trying to remove the lock.

I see that you don't put any ExpiresAt on the WriterQueue, leaving it null - they will hold forever (unless there is something else that causes it to crash out).  Would this be plausible for the WriterQueue?
* Have a timeout refresh interval of 10 seconds
* Set the ExpiresAt interval to 31 seconds
* Every time we try to acquire the writer lock and fail, we upsert the queued item with a new ExpiresAt time (31 seconds later)
* When we want to check for nothing queued, we look for non-expired items in the WriterQueue.

Is that overkill?  Should we try for this fix and see if it resolves it?

ddaspit (commented):
There are circumstances where the writer queue doesn't get cleared properly. Your proposed solution would definitely work. It would result in extra queries/updates to the database. We would be trying to acquire the lock every 10 seconds instead of waiting until the currently acquired lock expired. Also, we would need to do an extra update for the `ExpiresAt` property. Another option would be to add a timeout when attempting to acquire a lock. The timeout value could be used to set the `ExpiresAt` property for the writer queue entry. If we want to deal with this possible case, we can do it in another PR.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/491#-:-O6vcqeT7ewtZKGSHIwH:b-dwn52g)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

johnml1135:
Sure.

Issue #522, opened 2024-10-29
Get pretranslations for parallel corpora


Enkidu93:
The pretranslation service does not handle the new parallel corpora. This needs to be fixed ASAP. Also, there's the added complexity of multiple potential sources (i.e., where do we get the usfm from - always the first source?).

johnml1135:
Would this fix it #525?

Enkidu93:
The USFM portion of retrieving pretranslations still does not work like Peter noticed.

johnml1135:
The first portion was fixed.  More do to in other issues.

PR #529, opened 2024-10-31
Fix up USFM pretranslations for Parallel corpus

johnml1135:
#522

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/529)

### DIFF
--- a/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs
@@ -139,12 +139,16 @@ public class PreprocessBuildJob : HangfireBuildJob<IReadOnlyList<ParallelCorpus>
                     );
                 })
                 .ToArray();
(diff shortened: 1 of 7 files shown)


johnml1135:
*[`src/Serval/src/Serval.Translation/Services/PretranslationService.cs` line 55 at r1](https://reviewable.io/reviews/sillsdev/serval/529#-OAY1BOC02qt0aHXj9gw:-OAZElrpCOjtmQRDQBWt:b-1315om) ([raw file](https://github.com/sillsdev/serval/blob/46b128eb8a041b645dea77eabd3db896cd8b3822/src/Serval/src/Serval.Translation/Services/PretranslationService.cs#L55)):*

What about a situation in which we're using a parallel corpus that has two source (or target) corpora: one that includes MAT & MRK and another that includes MRK & JHN. Won't be run into issues if we try and insert pretranslations into the USFM for JHN from a project that isn't the one that contains JHN?
</blockquote></details>

Resolved through the meetings we had.

Issue #524, opened 2024-10-29
Back off of ClearML with "TooManyRequests"


johnml1135:
us polly or similar.

PR #531, opened 2024-10-31
retry clearml calls on 429 responses.

johnml1135:
https://github.com/sillsdev/serval/issues/524

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/531)

### DIFF
--- a/src/Machine/src/Serval.Machine.Shared/Configuration/IMachineBuilderExtensions.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Configuration/IMachineBuilderExtensions.cs
@@ -1,4 +1,5 @@
-﻿using Serval.Translation.V1;
+﻿using Polly.Extensions.Http;
+using Serval.Translation.V1;
 
 namespace Microsoft.Extensions.DependencyInjection;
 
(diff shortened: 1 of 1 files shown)


ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/531#-:-OAYs3og3sagHBUXKMs8:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Enkidu93 (changes requested):
Thank you for doing this, John! Have we figured out with ClearML if we're actually over-requesting because of an error in our code? Also, might it be wise to see if them upping the limit fixes things before introducing another change like this (i.e., this might increase the number of requests even while they're upping our limit)?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/531#-:-OAYrrdp6W9QnqIzgckb:b-ryikkn)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

johnml1135:
They just removed the limit entirely.  The issue may be that all of our servers present themselves as the same IP address.  Also, each agent calls out to ClearML every 2 seconds.  It's a number of things together.

johnml1135:
Moreover, we were intending the code to handle the 429 case and it was a mistake that it didn't.  That is why I am ok going forward with this change.

Issue #547, opened 2024-11-27
The parallel Corpus database holds onto old Filename when the file is updated


johnml1135:
It is copied to Corpus ... and then to Parallel Corpus ... and then the file is updated (and the filename changes) and then - poof!  it stops working!

johnml1135:
Option 1: when updating a file, after the new file is written, replace the old one with a link to the new one.
Option 2: Redo the data structures to actually reference the main DataFile, not their own copy.
Option 3: Keep track of where each file is used and when it is updated, update all of the other places it is used as well.

@ddaspit and @Enkidu93   - what do you think?  1 is the simplest and 2 is the most elegant.

johnml1135:
OPtion 2: and call it DataFileReference, not DataFile.

ddaspit:
We should use option 1: publish a data file updated message to the mediator. See `DataFileDeleted` for an example.

ddaspit:
Now that I looked at the code a bit more. I see that there are actually two issues.

1. In Serval.DataFiles, the `CorpusFile.File` property is a `DataFile`. That is incorrect. Unfortunately, I missed it in the code reviews. It should hold the ref id to the `DataFile` instance. The property should be renamed to `FileRef` and take a string. This is John's option 2.
2. When the data file is updated, the corresponding `CorpusFile` in Serval.Translation is not updated. We should use John's option 1 to fix this. We publish a data file updated message to the mediator, just like we do when a data file is deleted.

johnml1135:
@ddaspit, what do you think of the solution that I coded? I did a complete option 2.  There may be a cleaner way to do this, but I think that having one source of Truth for the file name makes the most sense to me.

PR #548, opened 2024-11-28
Don't store data filename in corpus data structure

johnml1135:
A fix for https://github.com/sillsdev/serval/issues/547.

Tests have not been updated yet - wanting to get review of approach.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/548)


ddaspit (changes requested):
> Reviewed 31 of 31 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/548)* status: all files reviewed, 5 unresolved discussions (waiting on @johnml1135)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 525 at r1](https://reviewable.io/reviews/sillsdev/serval/548#-OCnGTOKDMiehbDoMBMM:-OCnGTOKDMiehbDoMBMN:b19b5c3) ([raw file](https://github.com/sillsdev/serval/blob/ede3bb689dc45c7faa41214857f80d0f6ea31c12/src/Serval/src/Serval.Translation/Services/EngineService.cs#L525)):*
> ```Smalltalk
>     }
> 
>     public Task DeleteAllCorpusFilesAsync(string dataFileId, CancellationToken cancellationToken = default)
> ```

I noticed that this method was never updated to reflect the new corpus model changes. We should fix this.
___
*[`src/Serval/src/Serval.Shared/Contracts/CorpusResult.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/serval/548#-OCnA16h0GK2WDK6hdDK:-OCnA16h0GK2WDK6hdDL:b-vkxp4e) ([raw file](https://github.com/sillsdev/serval/blob/ede3bb689dc45c7faa41214857f80d0f6ea31c12/src/Serval/src/Serval.Shared/Contracts/CorpusResult.cs#L8)):*
> ```Smalltalk
>     public required string Language { get; init; }
>     public string? Name { get; init; }
>     public required IReadOnlyList<CorpusFile> Files { get; set; }
> ```

We don't want to pass models between contexts. We should only be passing contracts. A `CorpusFileResult` should be used to pass file information.
___
*[`src/Serval/src/Serval.DataFiles/Models/CorpusFile.cs` line 5 at r1](https://reviewable.io/reviews/sillsdev/serval/548#-OCnFmKwDqzrZpAJCjrs:-OCnFmKwDqzrZpAJCjrt:b4ems58) ([raw file](https://github.com/sillsdev/serval/blob/ede3bb689dc45c7faa41214857f80d0f6ea31c12/src/Serval/src/Serval.DataFiles/Models/CorpusFile.cs#L5)):*
> ```Smalltalk
> public record CorpusFile
> {
>     public required DataFileReference FileReference { get; init; }
> ```

I would prefer not to denormalize the database within a context. We shouldn't keep copies of the data. We also don't want clients to have to specify the same information twice. This should just be a `FileRef` property.
___
*[`src/Serval/src/Serval.Shared/Models/CorpusFile.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/serval/548#-OCnFr8xDCc5lSnfXJNM:-OCnFr8xDCc5lSnfXJNN:b-xb5jz4) ([raw file](https://github.com/sillsdev/serval/blob/ede3bb689dc45c7faa41214857f80d0f6ea31c12/src/Serval/src/Serval.Shared/Models/CorpusFile.cs#L9)):*
> ```Smalltalk
>     public required string TextId { get; set; }
> 
>     private string? _filename;
> ```

I see where you are going by not storing the filename and then synchronously calling Serval.DataFiles to get the filename when it is used. Rather than redesigning all of this, I would prefer that we stick with the current design. We are already dealing with one type of data file change, specifically deletion. We should just use the same approach when a file is updated. I would also like to keep the number of synchronous calls across contexts down to a minimum. In the current design, we only make these calls on setup. I would prefer not to make them on more frequent calls, such as starting a build or generating USFM. See the `DataFileDeletedConsumer` for an example.
___
*[`src/Serval/src/Serval.Shared/Models/MonolingualCorpus.cs` line 3 at r1](https://reviewable.io/reviews/sillsdev/serval/548#-OCnAHjo6Kk9LDnC0WZo:-OCnAHjo6Kk9LDnC0WZp:bfo6yx0) ([raw file](https://github.com/sillsdev/serval/blob/ede3bb689dc45c7faa41214857f80d0f6ea31c12/src/Serval/src/Serval.Shared/Models/MonolingualCorpus.cs#L3)):*
> ```Smalltalk
> namespace Serval.Shared.Models;
> 
> public record MonolingualCorpus
> ```

This should be moved back to Serval.Translation.

johnml1135:
Let me see if I understand the basic changes you are suggesting:
* Make a DataFileUpdated consumer - that updates:
  * The Corpus (as in the corpus controller)
  * The Legacy corpus
  * The Parallel Corpus
* Make a CorpusFileUpdated consumer that updates the Parallel Corpus (when the files change)

Issue #642, opened 2025-03-10
Index-Out-of-Range Error in `PretranslationService`


Enkidu93:
We've had a few of these in the Rancher alerts:
```
An unhandled exception was thrown by the application.
      System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
         at System.Collections.Generic.List`1.get_Item(Int32 index)
         at Serval.Translation.Services.PretranslationService.GetUsfmAsync(String engineId, Int32 modelRevision, String corpusId, String textId, PretranslationUsfmTextOrigin textOrigin, PretranslationUsfmTemplate template, CancellationToken cancellationToken) in /app/src/Serval/src/Serval.Translation/Services/PretranslationService.cs:line 53
         at Serval.Translation.Controllers.TranslationEnginesController.GetPretranslatedUsfmAsync(String id, String corpusId, String textId, Nullable`1 textOrigin, Nullable`1 template, CancellationToken cancellationToken) in /app/src/Serval/src/Serval.Translation/Controllers/TranslationEnginesController.cs:line 891
```

It looks like either a parallel corpus didn't have a source or target corpus or a monolingual subcorpus didn't have any files. It's easy enough to put a check here, but I'm going to try to see what exactly the situation is. Seems like maybe this reflects a bug/issue earlier on in the process.

Enkidu93:
This seems to have only happened in test team builds. I'll reach out to Mike and see if he's first-hand experienced this.

Enkidu93:
This happened again. I was wrong about the line. It's actually throwing this error when a deprecated corpus has no source files. (Although I imagine a similar issue could happen with parallel corpora). I suspect that someone deleted all the data files that were associated with a corpus. I'm not sure what the most appropriate way is of handling this. I think we could:
* Throw an error if a corpus with no source or target files is referenced in a build
* Throw a more informative error if a corpus with no source files is reference when fetching pretranslations
I think both of these circumstances would only result by accident, so it would be best to treat them as errors and inform the user.

Thoughts, @ddaspit @johnml1135?

ddaspit:
I agree that we should throw errors. Would it make sense to delete the corpus once all of its files are removed?

Enkidu93:
> I agree that we should throw errors. Would it make sense to delete the corpus once all of its files are removed?

[...]

PR #656, opened 2025-03-27
Fix empty corpora errors

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/642

Putting in a draft PR so in case I'm out for the baby's arrival, my progress on this task isn't lost if someone else needs to finish it. The thing that still needs to be added are a few more tests to cover all the checks I added (assuming you all don't have any issues with them).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/656)

### DIFF
--- a/src/Serval/src/Serval.Translation/Controllers/TranslationEnginesController.cs
+++ b/src/Serval/src/Serval.Translation/Controllers/TranslationEnginesController.cs
@@ -1353,6 +1353,12 @@ public class TranslationEnginesController(
             >(new GetCorpus { CorpusId = corpusId, Owner = Owner }, cancellationToken);
             if (response.Is(out Response<CorpusResult>? result))
             {
+                if (result.Message.Files.Any())
+                {
+                    throw new InvalidOperationException(
+                        $"The corpus {corpusId} does not have any files associated with it."
+                    );
+                }
                 corpora.Add(
                     new MonolingualCorpus
                     {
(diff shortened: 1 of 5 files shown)


ddaspit (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/656)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/Serval/src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 1435 at r1](https://reviewable.io/reviews/sillsdev/serval/656#-OMNNKxI2fOtSTOGIoCZ:-OMNNKxI2fOtSTOGIoC_:bna180e) ([raw file](https://github.com/sillsdev/serval/blob/45384fc08ad2437443d6af301b8fcaa10e52df1f/src/Serval/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L1435)):*
> ```Smalltalk
>                     );
>                 }
>                 Corpus corpus = engine.Corpora.Where(c => c.Id == pcc.CorpusId).First();
> ```

This can be simplified to a single `First` call with a predicate. Also, you should consider using `Single` instead of `First` if it is appropriate.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/656)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/Serval/src/Serval.Translation/Controllers/TranslationEnginesController.cs` line 1435 at r1](https://reviewable.io/reviews/sillsdev/serval/656#-OMNNKxI2fOtSTOGIoCZ:-OMNUfXYErZmJ8wdY6VE:bsucery) ([raw file](https://github.com/sillsdev/serval/blob/45384fc08ad2437443d6af301b8fcaa10e52df1f/src/Serval/src/Serval.Translation/Controllers/TranslationEnginesController.cs#L1435)):*

OK, I originally was using `First` because I assumed it was faster & it's definitely cleaner, but seems like, counterintuitively, using `Where().First()` is faster than using `First(predicate)` from what I saw online. It probably doesn't make much of a difference here though anyways, but I wonder what the best general practice is :thinking:. But yes, `Single` would definitely be more appropriate. Thank you!
</blockquote></details>

Generally, the best practice is to use the cleaner solution over the slightly faster solution unless there is a measured performance issue in the code.

ddaspit (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/656)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Enkidu93:
There are still a few more tests needed. I'll try and finish them today. I got side-tracked running a USFM-handling change example for Steven.

PR #659, opened 2025-03-30
Added disclaimer to the drafts

mudiagaobrikisil:
This PR adds a disclaimer line on the generated draft. The draft is pf the format
\rem This draft of MAT was machine translated on <date> from project <source_project_name>. It should be reviewed and edited carefully.
It adds a new property called addDisclaimer to the getUsfmAsync method, which is used in the translation engines controller class. This property, when set to true, ensures the disclaimer is shown and the other way round.
It also has two new tests to ensure the disclaimer is not added when the addDisclaimer is set to false, and that is is shown when the addDisclaimer is set to true.

The model name has not yet been added to the disclaimer. After discussing with @Enkidu93, it seemed there is no direct way to get the model name in serval, as it is in machine.py. It is captured in the settings.yaml of machine.py, e.g. parent_model_name: facebook/nllb-200-distilled-1.3B. The two suggestions were to either add a new variable to serval settings that captures the model name or to figure out a way to get it from clearml. However, I think the model name I have seen in ClearML is the one used for testing, not the actual model name. @johnml1135, @ddaspit, how would you suggest I go about the model name, and is it something I should do at this point?

Also, it has been manually tested on swagger and seems to work.


ddaspit (changes requested):
It might be useful to implement the insertion of the `\rem` marker in the USFM updater class, then silnlp would be able to use the same code.

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/659#-:-OMh7c7I1jAQFa9ewDk4:b-87mq5d)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/Serval/src/Serval.Translation/Services/IPretranslationService.cs` line 23 at r1](https://reviewable.io/reviews/sillsdev/serval/659#-OMgyuHJ6wkjhKVe8nj0:-OMgyuHJ6wkjhKVe8nj1:b-3d2692) ([raw file](https://github.com/sillsdev/serval/blob/f02950cf403a4725baf555693639f9b85a608b53/src/Serval/src/Serval.Translation/Services/IPretranslationService.cs#L23)):*
> ```Smalltalk
>         PretranslationUsfmMarkerBehavior embedBehavior,
>         PretranslationUsfmMarkerBehavior styleMarkerBehavior,
>         bool addDisclaimer = false,
> ```

We don't need this parameter.
___
*[`src/Serval/src/Serval.Translation/Services/PretranslationService.cs` line 184 at r1](https://reviewable.io/reviews/sillsdev/serval/659#-OMh74TqB-qBfNsYrv3F:-OMh74TqB-qBfNsYrv3G:b-t022o1) ([raw file](https://github.com/sillsdev/serval/blob/f02950cf403a4725baf555693639f9b85a608b53/src/Serval/src/Serval.Translation/Services/PretranslationService.cs#L184)):*
> ```Smalltalk
>             if (addDisclaimer)
>             {
>                 idPosition = usfm.IndexOf("\\id");
> ```

We should turn this into a function. Also, I believe that silnlp inserts the `\rem` after the `\ide` marker if it exists.

mudiagaobrikisil (commented):
Noted. Refactoring.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/659#-:-OMlKHft1OxI21mLoe9q:bq2qy6l)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

johnml1135:
If the model name is important enough for the disclaimer, it is also important enough for the build itself.  What if we put it in executionData and then grab it from there.  I would say open a new issue to grab the model name and put it in executionData but for right now, don't include it for this version of the disclaimer.

Enkidu93:
Let me know if you need help, Mudi, with reworking this. Also, @johnml1135 and I discussed the issue of adding the 'model name' to the remark this morning. For context, the 'model name' in EITL drafts includes the name of the base model, names of source project(s) and the target project, and the book filter that was applied to the data in training (e.g., NT+GEN) - so maybe 'model name' isn't really the best moniker :laughing: . John suggested that we could just try and get a basic disclaimer through and then spin off issues for getting the base model name and training data/filters specified. How does that sound, @ddaspit?

Enkidu93 (commented):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/659)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135 and @mudiagaobrikisil)

mudiagaobrikisil (commented):
@johnml1135, noted. I will refactor and get this disclaimer out and then open an issue for the model name. @Enkidu93 , would a call in the morning be fine?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/659#-:-OMm5FRz4Gjab0tBZxRu:b2ps87s)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135)

ddaspit (commented):
@Enkidu93 I agree. We should just start with a basic disclaimer.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/659#-:-OMmjY64E7JZ24hDVtqX:bfl51mz)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135 and @mudiagaobrikisil)

Enkidu93:
Since the strategy that we've arrived at for implementing this is different than what Mudi coded, I'm just going to close this and open a new PR.

Issue #685, opened 2025-05-12
Update source/target language endpoint is broken


Enkidu93:
Work [here](https://github.com/sillsdev/serval/pull/624) allowed for the source and target languages of an engine to be updated. The endpoint will update codes on the Serval side but not on the machine engine side, meaning that outdated language codes will be used for builds.

PR #686, opened 2025-05-12
Update language codes in the machine engine as well as Serval

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/685

See comments in code. I wasn't sure what would be appropriate when the language codes are changed - should we delete pretranslations or cancel any current builds? I'm thinking we ought to, but will be need to coordinate more with SF on this then?

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/686)

### DIFF
--- a/src/Machine/src/Serval.Machine.Shared/Services/ITranslationEngineService.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/ITranslationEngineService.cs
@@ -14,6 +14,13 @@ public interface ITranslationEngineService
     );
     Task DeleteAsync(string engineId, CancellationToken cancellationToken = default);
 
+    Task UpdateAsync(
+        string engineId,
+        string sourceLanguage,
+        string targetLanguage,
+        CancellationToken cancellationToken = default
+    );
+
     Task<IReadOnlyList<TranslationResult>> TranslateAsync(
         string engineId,
         int n,
--- a/src/Machine/src/Serval.Machine.Shared/Services/NmtEngineService.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/NmtEngineService.cs
@@ -65,6 +65,26 @@ public class NmtEngineService(
         await _buildJobService.DeleteEngineAsync(engineId, CancellationToken.None);
     }
 
+    public async Task UpdateAsync(
+        string engineId,
+        string sourceLanguage,
+        string targetLanguage,
+        CancellationToken cancellationToken = default
+    )
+    {
+        // await CancelBuildJobAsync(engineId, cancellationToken); Should we cancel the build?
+
+        await _engines.UpdateAsync(
+            e => e.EngineId == engineId,
+            u =>
+            {
+                u.Set(e => e.SourceLanguage, sourceLanguage);
+                u.Set(e => e.TargetLanguage, targetLanguage);
+            },
+            cancellationToken: cancellationToken
+        );
+    }
+
     public async Task StartBuildAsync(
         string engineId,
         string buildId,
--- a/src/Machine/src/Serval.Machine.Shared/Services/NmtEngineService.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/NmtEngineService.cs
@@ -72,7 +72,7 @@ public class NmtEngineService(
         CancellationToken cancellationToken = default
     )
     {
-        // await CancelBuildJobAsync(engineId, cancellationToken); Should we cancel the build?
+        await CancelBuildJobAsync(engineId, cancellationToken);
 
         await _engines.UpdateAsync(
             e => e.EngineId == engineId,
--- a/src/Machine/src/Serval.Machine.Shared/Services/ServalTranslationEngineServiceV1.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/ServalTranslationEngineServiceV1.cs
@@ -33,6 +33,18 @@ public class ServalTranslationEngineServiceV1(IEnumerable<ITranslationEngineServ
         return Empty;
     }
 
+    public override async Task<Empty> Update(UpdateRequest request, ServerCallContext context)
+    {
+        ITranslationEngineService engineService = GetEngineService(request.EngineType);
+        await engineService.UpdateAsync(
+            request.EngineId,
+            request.SourceLanguage,
+            request.TargetLanguage,
+            context.CancellationToken
+        );
+        return Empty;
+    }
+
     public override async Task<TranslateResponse> Translate(TranslateRequest request, ServerCallContext context)
     {
         ITranslationEngineService engineService = GetEngineService(request.EngineType);
--- a/src/Machine/src/Serval.Machine.Shared/Services/SmtTransferEngineService.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/SmtTransferEngineService.cs
@@ -85,6 +85,26 @@ public class SmtTransferEngineService(
         await _lockFactory.DeleteAsync(engineId, CancellationToken.None);
     }
 
+    public async Task UpdateAsync(
+        string engineId,
+        string sourceLanguage,
+        string targetLanguage,
+        CancellationToken cancellationToken = default
+    )
+    {
+        // await CancelBuildJobAsync(engineId, cancellationToken); Should we cancel the build or delete the segment pairs?
+
+        await _engines.UpdateAsync(
+            e => e.EngineId == engineId,
+            u =>
+            {
+                u.Set(e => e.SourceLanguage, sourceLanguage);
+                u.Set(e => e.TargetLanguage, targetLanguage);
+            },
+            cancellationToken: cancellationToken
+        );
+    }
+
     public async Task<IReadOnlyList<TranslationResult>> TranslateAsync(
         string engineId,
         int n,
--- a/src/Machine/test/Serval.Machine.Shared.Tests/Services/NmtEngineServiceTests.cs
+++ b/src/Machine/test/Serval.Machine.Shared.Tests/Services/NmtEngineServiceTests.cs
@@ -66,6 +66,16 @@ public class NmtEngineServiceTests
         Assert.That(env.Engines.Contains("engine1"), Is.False);
     }
 
+    [Test]
+    public async Task UpdateAsync()
+    {
+        using var env = new TestEnvironment();
+        await env.Service.UpdateAsync("engine1", "fr", "en");
+        TranslationEngine engine = env.Engines.Get("engine1");
+        Assert.That(engine.SourceLanguage, Is.EqualTo("fr"));
+        Assert.That(engine.TargetLanguage, Is.EqualTo("en"));
+    }
+
     private class TestEnvironment : DisposableBase
     {
         private readonly Hangfire.InMemory.InMemoryStorage _memoryStorage;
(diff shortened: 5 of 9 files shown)


ddaspit (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/686)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 215 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ44CCM0F8FtsVim-JI:-OQ44CCM0F8FtsVim-JJ:bv6uy9y) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Serval/src/Serval.Translation/Services/EngineService.cs#L215)):*
> ```Smalltalk
>             targetLanguage = engine.TargetLanguage;
> 
>         TranslationEngineApi.TranslationEngineApiClient client =
> ```

We should perform the update to the local DB first in a transaction. That way if the gRPC call fails, Serval will rollback the update and not leave things in an inconsistent state.
___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 237 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ45hKtB-9WO9kaZxpl:-OQ45hKu8_XvdCrTcSih:b-8q2w5q) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Serval/src/Serval.Translation/Services/EngineService.cs#L237)):*
> ```Smalltalk
>             cancellationToken: CancellationToken.None
>         );
>         // await _pretranslations.DeleteAllAsync(pt => pt.EngineRef == engineId, cancellationToken); Should we delete the pretranslations?
> ```

Yes, let's delete all pretranslations. When we update an engine, we are resetting it.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtEngineService.cs` line 75 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ45YQZAKnP-N-CYQJM:-OQ45YQZAKnP-N-CYQJN:b-i7pgd1) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Machine/src/Serval.Machine.Shared/Services/NmtEngineService.cs#L75)):*
> ```Smalltalk
>     )
>     {
>         // await CancelBuildJobAsync(engineId, cancellationToken); Should we cancel the build?
> ```

Yes, let's go ahead and cancel the current build.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/686)* status: 6 of 8 files reviewed, 3 unresolved discussions (waiting on @ddaspit)

___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtEngineService.cs` line 75 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ45YQZAKnP-N-CYQJM:-OQ4Apbz59Zx7GQ35OJK:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Machine/src/Serval.Machine.Shared/Services/NmtEngineService.cs#L75)):*

Yes, let's go ahead and cancel the current build.
</blockquote></details>

Done.
___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 215 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ44CCM0F8FtsVim-JI:-OQ4BZ7vCjMhmtwpFPtG:b-4zdnti) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Serval/src/Serval.Translation/Services/EngineService.cs#L215)):*

We should perform the update to the local DB first in a transaction. That way if the gRPC call fails, Serval will rollback the update and not leave things in an inconsistent state.
</blockquote></details>

Done. Why is it the reverse in `DeleteAsync`? Wouldn't we want to call gRPC first so that if it fails, we don't update the DB at all? Or am I not understanding how the data access context works?
___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 237 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ45hKtB-9WO9kaZxpl:-OQ4BYrr0qR_Zcyj-UUT:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Serval/src/Serval.Translation/Services/EngineService.cs#L237)):*

Yes, let's delete all pretranslations. When we update an engine, we are resetting it.
</blockquote></details>

Done.

Enkidu93:
How easy is it to change language codes on your project in SF? My only concern with cancelling builds and deleting pretranslations is that if this is something a user can do easily, they might accidentally change them mid build or while working with drafts and not realize they're going to lose data. Can they only change language codes at training/drafting time?

ddaspit (commented):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/686)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 215 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ44CCM0F8FtsVim-JI:-OQ4s3St3IgLex6V4mxZ:byc26hc) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Serval/src/Serval.Translation/Services/EngineService.cs#L215)):*

Done. Why is it the reverse in `DeleteAsync`? Wouldn't we want to call gRPC first so that if it fails, we don't update the DB at all? Or am I not understanding how the data access context works?
</blockquote></details>

That is a good question. The consistency model between Serval and the engine services is tricky, because we use synchronous calls (gRPC).  For creation, we use the following pattern (something like a simplified Saga pattern):
1. Serval creates the entity with `IsInitialized` set to `false`.
2. Serval asks the engine service to create the entity.
3. Serval receives a response from the engine:
a. If unsuccessful, rollback the entity creation.
b. If successful, set `IsInitalized` to `true`.
c. If there is a catastrophic failure (Serval or DB goes down), the entity gets cleaned up later by a background service.

We never updated `DeleteAsync` to use a similar pattern, but we should. We can just set `IsInitialized` to `false` before we ask the engine service to delete the engine.

For now, I was thinking we could do something simple to get us partial consistency guarantees for the update case, so that we can get this fixed quickly. A "correct" solution would be similar to the pattern that we use when we create an entity.

We really need to implement a proper consistency model (Saga-like) that we can apply to all calls like this, but that would take more time.

Also, the gRPC call should occur inside of the transaction, so that if it fails, the transaction is rolled back.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/686)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 215 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ44CCM0F8FtsVim-JI:-OQ5UgPsELoQgSjfKYjb:b-a8m5hp) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Serval/src/Serval.Translation/Services/EngineService.cs#L215)):*

That is a good question. The consistency model between Serval and the engine services is tricky, because we use synchronous calls (gRPC).  For creation, we use the following pattern (something like a simplified Saga pattern):
1. Serval creates the entity with `IsInitialized` set to `false`.
2. Serval asks the engine service to create the entity.
3. Serval receives a response from the engine:
a. If unsuccessful, rollback the entity creation.
b. If successful, set `IsInitalized` to `true`.
c. If there is a catastrophic failure (Serval or DB goes down), the entity gets cleaned up later by a background service.

We never updated `DeleteAsync` to use a similar pattern, but we should. We can just set `IsInitialized` to `false` before we ask the engine service to delete the engine.

For now, I was thinking we could do something simple to get us partial consistency guarantees for the update case, so that we can get this fixed quickly. A "correct" solution would be similar to the pattern that we use when we create an entity.

We really need to implement a proper consistency model (Saga-like) that we can apply to all calls like this, but that would take more time.

Also, the gRPC call should occur inside of the transaction, so that if it fails, the transaction is rolled back.
</blockquote></details>

That's very helpful - thank you. I spun off https://github.com/sillsdev/serval/issues/689 to capture the issue with `DeleteAsync`. Let me know if you'd like me to just go ahead and do that in this PR too.

ddaspit (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/686#-:-OQ5bx6Y461xaID9cvWJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 215 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ44CCM0F8FtsVim-JI:-OQ5bQ5O0q10CLsblS05:b-iad9qf) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Serval/src/Serval.Translation/Services/EngineService.cs#L215)):*

That's very helpful - thank you. I spun off https://github.com/sillsdev/serval/issues/689 to capture the issue with `DeleteAsync`. Let me know if you'd like me to just go ahead and do that in this PR too.
</blockquote></details>

There are a couple of other changes that we will need to make to properly support `DeleteAsync`. Let's just go ahead and get this fix in. We can fix `DeleteAsync` later.

Enkidu93 (commented):
Did you see Michael's comment [here](https://sil-language-software.slack.com/archives/C07PH69AC1F/p1747081420512829?thread_ts=1747008713.022109&cid=C07PH69AC1F)? I guess we should be fine because the language code is only updated when they change the source or target project in which case we'd really want the pretranslations to be deleted anyways, and because the draft is cached, it shouldn't be an issue. I'll just make sure to alert Peter to this change when this goes to QA and we can adjust need be. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/686#-:-OQ61een0XudZ2mnFzma:bsr207i)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 215 at r1](https://reviewable.io/reviews/sillsdev/serval/686#-OQ44CCM0F8FtsVim-JI:-OQ61cf767Bp_PsFAo60:b2nhktp) ([raw file](https://github.com/sillsdev/serval/blob/15f2847ee11bf58c8098a1576f9cd273a2666323/src/Serval/src/Serval.Translation/Services/EngineService.cs#L215)):*

There are a couple of other changes that we will need to make to properly support `DeleteAsync`. Let's just go ahead and get this fix in. We can fix `DeleteAsync` later.
</blockquote></details>

OK.

Issue #678, opened 2025-04-28
Take a Corpus Snapshot when Building


Enkidu93:
At the moment, pretranslations can be generated using one revision of a corpus and then that corpus can be revised, and when the USFM is generated, the newer revision will be used. In general, any processing connected to formatting or otherwise post-processing pretranslations ought to operate on the corpus as it was when the pretranslations were generated.

Spin off of #665.

ddaspit:
I think something like that would work okay. There might be an issue for projects that have both translation suggestions and AI drafting enabled. I believe that SF uses the same corpus for both the SMT and NMT engines for a single project. If the corpus gets updated for the SMT engine, it would invalidate the NMT pretranslations. We would need to run it by the SF team to make sure it wouldn't cause issues for them. I would prefer to delete the pretranslations when the corpus is updated, so the behavior is consistent (i.e. if a corpus is deleted or updated, the associated pretranslations are deleted).

In any case, I think it would be worth going through the exercise of designing the corpus snapshot solution, since it would provide calling applications with the most flexibility. We would only need to keep the corpus snapshot for the most recent build (a single snapshot per engine), since we only keep the pretranslations for the most recent build (if I remember correctly).

Enkidu93:
> I think something like that would work okay. There might be an issue for projects that have both translation suggestions and AI drafting enabled. I believe that SF uses the same corpus for both the SMT and NMT engines for a single project. If the corpus gets updated for the SMT engine, it would invalidate the NMT pretranslations. We would need to run it by the SF team to make sure it wouldn't cause issues for them. I would prefer to delete the pretranslations when the corpus is updated, so the behavior is consistent (i.e. if a corpus is deleted or updated, the associated pretranslations are deleted).
> 
> In any case, I think it would be worth going through the exercise of designing the corpus snapshot solution, since it would provide calling applications with the most flexibility. We would only need to keep the corpus snapshot for the most recent build (a single snapshot per engine), since we only keep the pretranslations for the most recent build (if I remember correctly).

During today's draft quality improvement meeting, Nathaniel gave us the go ahead to implement the simpler just-delete-them approach.

PR #687, opened 2025-05-12
Delete pretranslations when a corpus is updated

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/678

Sorry for the delay in getting this in.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/687)

### DIFF
--- a/src/DataAccess/src/SIL.DataAccess/MongoDataAccessContext.cs
+++ b/src/DataAccess/src/SIL.DataAccess/MongoDataAccessContext.cs
@@ -13,6 +13,10 @@ public class MongoDataAccessContext(IMongoClient client) : DisposableBase, IMong
     )
     {
         IClientSessionHandle session = await StartSession(cancellationToken).ConfigureAwait(false);
+        if (session.IsInTransaction)
+        {
+            return await callbackAsync(cancellationToken).ConfigureAwait(false);
+        }
 
         return await session
             .WithTransactionAsync(
--- a/src/Serval/src/Serval.Translation/Services/EngineService.cs
+++ b/src/Serval/src/Serval.Translation/Services/EngineService.cs
@@ -486,6 +486,9 @@ public class EngineService(
         );
         if (engine is null)
             throw new EntityNotFoundException($"Could not find the Corpus '{corpusId}' in Engine '{engineId}'.");
+
+        await _pretranslations.DeleteAllAsync(pt => pt.CorpusRef == corpusId, cancellationToken: cancellationToken);
+
         return engine.Corpora.First(c => c.Id == corpusId);
     }
 
@@ -571,6 +574,11 @@ public class EngineService(
             );
         }
 
+        await _pretranslations.DeleteAllAsync(
+            pt => pt.CorpusRef == parallelCorpusId,
+            cancellationToken: cancellationToken
+        );
+
         return engine.ParallelCorpora.First(c => c.Id == parallelCorpusId);
     }
 
@@ -602,9 +610,9 @@ public class EngineService(
         );
     }
 
-    public Task DeleteAllCorpusFilesAsync(string dataFileId, CancellationToken cancellationToken = default)
+    public async Task DeleteAllCorpusFilesAsync(string dataFileId, CancellationToken cancellationToken = default)
     {
-        return Entities.UpdateAllAsync(
+        await Entities.UpdateAllAsync(
             e =>
                 e.Corpora.Any(c =>
                     c.SourceFiles.Any(f => f.Id == dataFileId) || c.TargetFiles.Any(f => f.Id == dataFileId)
@@ -628,15 +636,44 @@ public class EngineService(
             },
             cancellationToken: cancellationToken
         );
+        HashSet<string> parallelCorpusIds = (
+            await Entities.GetAllAsync(
+                e =>
+                    e.ParallelCorpora.Any(c =>
+                        c.SourceCorpora.Any(cs => cs.Files.Any(f => f.Id == dataFileId))
+                        || c.TargetCorpora.Any(tc => tc.Files.Any(f => f.Id == dataFileId))
+                    ),
+                cancellationToken: cancellationToken
+            )
+        )
+            .SelectMany(e => e.ParallelCorpora.Select(c => c.Id))
+            .ToHashSet();
+
+        HashSet<string> corpusIds = (
+            await Entities.GetAllAsync(
+                e =>
+                    e.Corpora.Any(c =>
+                        c.SourceFiles.Any(f => f.Id == dataFileId) || c.TargetFiles.Any(f => f.Id == dataFileId)
+                    ),
+                cancellationToken
+            )
+        )
+            .SelectMany(e => e.Corpora.Select(c => c.Id))
+            .ToHashSet();
+
+        await _pretranslations.DeleteAllAsync(
+            pt => parallelCorpusIds.Contains(pt.CorpusRef) || corpusIds.Contains(pt.CorpusRef),
+            cancellationToken: cancellationToken
+        );
     }
 
-    public Task UpdateDataFileFilenameFilesAsync(
+    public async Task UpdateDataFileFilenameFilesAsync(
         string dataFileId,
         string filename,
         CancellationToken cancellationToken = default
     )
     {
-        return Entities.UpdateAllAsync(
+        await Entities.UpdateAllAsync(
             e =>
                 e.Corpora.Any(c =>
                     c.SourceFiles.Any(f => f.Id == dataFileId) || c.TargetFiles.Any(f => f.Id == dataFileId)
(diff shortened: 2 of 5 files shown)


ddaspit (changes requested):
> Reviewed 2 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/687)* status: 2 of 4 files reviewed, 3 unresolved discussions

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 471 at r1](https://reviewable.io/reviews/sillsdev/serval/687#-OQ5cbDW1of4LNzCoBkH:-OQ5cbDW1of4LNzCoBkI:b-vtwmul) ([raw file](https://github.com/sillsdev/serval/blob/c23bd12a2b472d418c6b58bd309e08ed95e4cf39/src/Serval/src/Serval.Translation/Services/EngineService.cs#L471)):*
> ```Smalltalk
>             throw new EntityNotFoundException($"Could not find the Corpus '{corpusId}' in Engine '{engineId}'.");
> 
>         await _pretranslations.DeleteAllAsync(pt => pt.CorpusRef == corpusId, cancellationToken: cancellationToken);
> ```

This should be performed in a transaction with the update call.
___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 558 at r1](https://reviewable.io/reviews/sillsdev/serval/687#-OQ5coYL5gW9NPrr53Da:-OQ5coYL5gW9NPrr53Db:b-vtwmul) ([raw file](https://github.com/sillsdev/serval/blob/c23bd12a2b472d418c6b58bd309e08ed95e4cf39/src/Serval/src/Serval.Translation/Services/EngineService.cs#L558)):*
> ```Smalltalk
>         }
> 
>         await _pretranslations.DeleteAllAsync(
> ```

This should be performed in a transaction with the update call.
___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 596 at r1](https://reviewable.io/reviews/sillsdev/serval/687#-OQ5t9auBQZ9NSAtEhCZ:-OQ5t9auBQZ9NSAtEhC_:b-ilc9xc) ([raw file](https://github.com/sillsdev/serval/blob/c23bd12a2b472d418c6b58bd309e08ed95e4cf39/src/Serval/src/Serval.Translation/Services/EngineService.cs#L596)):*
> ```Smalltalk
>     public async Task DeleteAllCorpusFilesAsync(string dataFileId, CancellationToken cancellationToken = default)
>     {
>         HashSet<string> parallelCorpusIds = (
> ```

These non-atomic reads/writes make me concerned about race conditions. If this is the only way to make it work, then we might need to fall back to the snapshot solution.

Enkidu93 (commented):
Separate from the long question below and more in regard to design (assuming we can use this approach at all), I wonder if instead of querying where the file is used here, we could do that in the `CorpusService` which is also being triggered by the `DataFileDeleted` being published and then push a `CorpusUpdated` there. That seems a little cleaner to me, but is there a reason I shouldn't do that? 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/687#-:-OQEKknoBgu0W7SLoIe3:b-6woamh)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @ddaspit)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 596 at r1](https://reviewable.io/reviews/sillsdev/serval/687#-OQ5t9auBQZ9NSAtEhCZ:-OQEFRYp570V72yIrndl:bwhmo2m) ([raw file](https://github.com/sillsdev/serval/blob/c23bd12a2b472d418c6b58bd309e08ed95e4cf39/src/Serval/src/Serval.Translation/Services/EngineService.cs#L596)):*

These non-atomic reads/writes make me concerned about race conditions. If this is the only way to make it work, then we might need to fall back to the snapshot solution.
</blockquote></details>

I've been stewing on this a bit. Sorry for the delay:

Isn't `UpdateAllAsync` itself a non-atomic read/write? If I understand correctly, Mongo scans with the filter and then updates, so it could have stale reads. Only `findAndUpdateOne` is atomic, right? (This isn't to say that we should add more opportunities for race conditions - but if this is a concern, maybe that ought to be as well). I don't think this is a concern though because the only race-condition situation I can think of is:
* (1) Operation A finds all corpora where fileId is used
* (2) Operation B updates a corpus to reference fileId
* (3) Operation A removes all references to fileId from the corpora it fetched earlier in (1)
* (4) Now there's a dangling reference to fileId in the corpus that B updated in (2)
But that isn't really an issue since this is all happening after the file with fileId1 has been deleted, right? So (2) would fail, wouldn't it? (Alternatively, (2) could *remove* a reference to fileId but that's also not an issue because it's going to get removed anyways).

Similar to the scenario above, I'm not sure that it matters that the pretranslation deletion/corpus filtering is non-atomic. If by the time the pretranslations are deleted, a corpus references the file and didn't before, that should have failed because the file would have been deleted already. And if by the time the pretranslations are deleted, a corpus has *removed* a reference to the file, it also shouldn't make a difference - the right thing to do would still to be to delete the pretranslations.

This is based on the assumption that this occurs after the file has been deleted. Is that same to assume or not? And if not, is there a way we can force that to be the case?

I might be way off. Let me know what you think. Assuming it is 'safe', I can still put these in transactions in case something fails.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/687)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @ddaspit)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 471 at r1](https://reviewable.io/reviews/sillsdev/serval/687#-OQ5cbDW1of4LNzCoBkH:-OQJO6UW8dkXV1_4k57F:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/c23bd12a2b472d418c6b58bd309e08ed95e4cf39/src/Serval/src/Serval.Translation/Services/EngineService.cs#L471)):*

This should be performed in a transaction with the update call.
</blockquote></details>

Done.
___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 558 at r1](https://reviewable.io/reviews/sillsdev/serval/687#-OQ5coYL5gW9NPrr53Da:-OQJUYOY3jRpoQv-EBv1:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/c23bd12a2b472d418c6b58bd309e08ed95e4cf39/src/Serval/src/Serval.Translation/Services/EngineService.cs#L558)):*

This should be performed in a transaction with the update call.
</blockquote></details>

Done.
___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 596 at r1](https://reviewable.io/reviews/sillsdev/serval/687#-OQ5t9auBQZ9NSAtEhCZ:-OQJbyUm4qrxgIr24LpS:b2jovtg) ([raw file](https://github.com/sillsdev/serval/blob/c23bd12a2b472d418c6b58bd309e08ed95e4cf39/src/Serval/src/Serval.Translation/Services/EngineService.cs#L596)):*

I've been stewing on this a bit. Sorry for the delay:

Isn't `UpdateAllAsync` itself a non-atomic read/write? If I understand correctly, Mongo scans with the filter and then updates, so it could have stale reads. Only `findAndUpdateOne` is atomic, right? (This isn't to say that we should add more opportunities for race conditions - but if this is a concern, maybe that ought to be as well). I don't think this is a concern though because the only race-condition situation I can think of is:
* (1) Operation A finds all corpora where fileId is used
* (2) Operation B updates a corpus to reference fileId
* (3) Operation A removes all references to fileId from the corpora it fetched earlier in (1)
* (4) Now there's a dangling reference to fileId in the corpus that B updated in (2)
But that isn't really an issue since this is all happening after the file with fileId1 has been deleted, right? So (2) would fail, wouldn't it? (Alternatively, (2) could *remove* a reference to fileId but that's also not an issue because it's going to get removed anyways).

Similar to the scenario above, I'm not sure that it matters that the pretranslation deletion/corpus filtering is non-atomic. If by the time the pretranslations are deleted, a corpus references the file and didn't before, that should have failed because the file would have been deleted already. And if by the time the pretranslations are deleted, a corpus has *removed* a reference to the file, it also shouldn't make a difference - the right thing to do would still to be to delete the pretranslations.

This is based on the assumption that this occurs after the file has been deleted. Is that same to assume or not? And if not, is there a way we can force that to be the case?

I might be way off. Let me know what you think. Assuming it is 'safe', I can still put these in transactions in case something fails.
</blockquote></details>

After discussing, we resolved to just throw these in a transaction, and that should be good enough to keep the data consistent even if there's a slim chance for some stale reads. And like I outlined above, the stale reads shouldn't make a big difference in these particular cases.

Enkidu93:
Working on addressing the failing tests

ddaspit (commented):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r2, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/687)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 646 at r4](https://reviewable.io/reviews/sillsdev/serval/687#-OQL0O842dWowGs0mZ1O:-OQL0O842dWowGs0mZ1P:b7e3zb4) ([raw file](https://github.com/sillsdev/serval/blob/e9e20af9c03859f99725e35ac70b9ae21e0358af/src/Serval/src/Serval.Translation/Services/EngineService.cs#L646)):*
> ```Smalltalk
>     public async Task DeleteAllCorpusFilesAsync(string dataFileId, CancellationToken cancellationToken = default)
>     {
>         //This is only getting called within a transaction
> ```

A caller might expect this method to be atomic, since all other methods use transactions. Right now, there is only one caller, which we know has created a transaction already. I'm worried that something else might call this method in the future with the expectation that it works like all other methods. I think we should make `WithTransactionAsync` reentrant. It shouldn't be difficult. A client session is created to manage transactions. The session object has a `IsInTransaction` property that you can check.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/687)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`src/Serval/src/Serval.Translation/Services/EngineService.cs` line 646 at r4](https://reviewable.io/reviews/sillsdev/serval/687#-OQL0O842dWowGs0mZ1O:-OQLKniIDowr-wSzDQst:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/e9e20af9c03859f99725e35ac70b9ae21e0358af/src/Serval/src/Serval.Translation/Services/EngineService.cs#L646)):*

A caller might expect this method to be atomic, since all other methods use transactions. Right now, there is only one caller, which we know has created a transaction already. I'm worried that something else might call this method in the future with the expectation that it works like all other methods. I think we should make `WithTransactionAsync` reentrant. It shouldn't be difficult. A client session is created to manage transactions. The session object has a `IsInTransaction` property that you can check.
</blockquote></details>

Done.

ddaspit (approved):
:lgtm:

> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/687#-:-OQOMr8786VdTraXt3v7:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Issue #695, opened 2025-05-21
Investigate issues with `clearml_stats.py` script


Enkidu93:
JoEllen recently flagged some unusual data coming out of this script - jobs are being misclassified as `research` and potentially jobs are being filtered or aggregated incorrectly. Instead of or in addition to processing the raw data and putting it into a purpose-specific spreadsheet like it is now, it might be helpful to also output the more 'raw' data (including engine and build ids) so that it can be post-processed as needed.

ddaspit:
This issue could be related to #696.

PR #706, opened 2025-06-20
Improved ClearML Stats

Enkidu93:
Changes include:
* Efficiency improvements: 3-4 hours to run down to 10-15 minutes.
* Improve language code fetching. (Previously, many rows were getting marked as 'unknown'). 
* Fix misidentified research jobs. (Internal QA jobs were getting marked as research because of a typo).
* Make script work when run the first time. (Previously, the code would only run properly if you'd run it previously on your machine).
* Aggregate over the ClearML project name to maintain project uniqueness per line. (Previously, production, research, and dev jobs were colliding as well as situations in which more than one project was working with the same minority language - e.g., FT and BT).
* Move to using a jupyter notebook for data exploration. 
* Add basic interactive choropleth plot to see where our service has been used. (Super neat :grin:). 

Future improvements not done here/takeaways:
* We should remove thot jobs from the listing or otherwise process them differently
* We should try and differentiate between SF QA and SF Prod jobs. I think we might actually be able to do this already with a hacky solution. Probably it's better to tag them in the ClearML API.
* There's likely more useful data we can extract from ClearML if we dig through the configs - lots of interesting opportunities here for retrospective research!
* I added a field to capture/explored language projects who've used Serval regularly and for a long time. This group might be a helpful bunch to reach out to about how SF has helped them. Also, we could potentially use these languages (a fairly diverse list!) as a research testing standard - that way, we'd know how our features affect frequent users. 

(I'd love to follow up/work on any and all of the above!)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/706)


Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/695

ddaspit (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/706#-:-OTT98Pm2XiawWo3dOAn:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

PR #718, opened 2025-08-05
Add quotation analysis to NMT builds

### LINKED ISSUE #665: Corpora - store and retrieve cachedAnalysis
To support punctuation de-normalization, we need to store data with the corpus.
* We can create a dict of string: strings called cachedAnalysis that can be read/writeable from the rest of Serval.
* With this update, other code in Serval can create, cache, and read analysis data as stored in the corpus.
* We will clear (delete) all cachedAnalysis when any file is updated on the corpus.

pmachapman:
Resolves #665.

Requires sillsdev/machine/pull/316.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/718)


pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/718)* status: all files reviewed, 4 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`src/Machine/src/Serval.Machine.Shared/Services/TranslationPreprocessBuildJob.cs` line 124 at r1](https://reviewable.io/reviews/sillsdev/serval/718#-OX-T8CN3AelFBLyHAAK:-OX0LkUqDWCUrJ23OBum:b-m7pgq7) ([raw file](https://github.com/sillsdev/serval/blob/57f706ddee2d33031cbbc052d1029fbf69323b58/src/Machine/src/Serval.Machine.Shared/Services/TranslationPreprocessBuildJob.cs#L124)):*

This is only needed for NMT jobs. I think this should be implemented in `NmtPreprocessBuildJob`.
</blockquote></details>

Done. To do this, I needed to change the base method to be `virtual` instead of `abstract.
___
*[`src/Serval/src/Serval.Grpc/Protos/serval/translation/v1/platform.proto` line 77 at r1](https://reviewable.io/reviews/sillsdev/serval/718#-OX-VBxQ40sQlX5l_YGH:-OX0PMHa-NdDJAiz8rc2:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/57f706ddee2d33031cbbc052d1029fbf69323b58/src/Serval/src/Serval.Grpc/Protos/serval/translation/v1/platform.proto#L77)):*

I would rename this and the associated result class to include `Parallel` so that it is clear that this is referring to a parallel corpus and not a monolingual corpus.
</blockquote></details>

Done.
___
*[`src/Serval/src/Serval.Shared/Models/CorpusAnalysis.cs` line 5 at r1](https://reviewable.io/reviews/sillsdev/serval/718#-OX-UA_c-jlKk3v8KFMW:-OX0PN2w1tzG4on7oluw:b-896fiz) ([raw file](https://github.com/sillsdev/serval/blob/57f706ddee2d33031cbbc052d1029fbf69323b58/src/Serval/src/Serval.Shared/Models/CorpusAnalysis.cs#L5)):*

This should be `ParallelCorpusRef` so that the naming is consistent with the `PretranslateCorpus` model. Do we need to continue to support the obsolete `Corpus` data model with this new feature? I would prefer not to.
</blockquote></details>

Done,
___
*[`src/ServiceToolkit/src/SIL.ServiceToolkit/Services/ParallelCorpusPreprocessingService.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/serval/718#-OX-X6Ke1HKqv6rJBD4T:-OX0dVk_0l0oc0RkCKd1:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/57f706ddee2d33031cbbc052d1029fbf69323b58/src/ServiceToolkit/src/SIL.ServiceToolkit/Services/ParallelCorpusPreprocessingService.cs#L8)):*

Could this method just return the conventions rather than accept a delegate? We could just run this on a single corpus to simplify the return value.
</blockquote></details>

Done.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/718)* status: 39 of 43 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`src/Serval/src/Serval.Translation/Services/TranslationPlatformServiceV1.cs` line 305 at r2](https://reviewable.io/reviews/sillsdev/serval/718#-OXA1G0nAUPw7tqlVelZ:-OXLQYCA4iURHDmgrr_e:b-ioxi27) ([raw file](https://github.com/sillsdev/serval/blob/d33e0326aa4e411d69b9309bec668ba90facbcab/src/Serval/src/Serval.Translation/Services/TranslationPlatformServiceV1.cs#L305)):*

I'm not sure exactly what happens if no update ops are called. It might result in an unnecessary Mongo call. Just to be safe, it would make sense to wrap the entire `UpdateAsync` call in this check.
</blockquote></details>

Done. Thank you!
___
*[`src/Serval/src/Serval.Translation/Services/TranslationPlatformServiceV1.cs` line 312 at r2](https://reviewable.io/reviews/sillsdev/serval/718#-OXA1g2j0boLeYnpVL-1:-OXLQuj4-ScT0HYpiW2y:b2qvpa4) ([raw file](https://github.com/sillsdev/serval/blob/d33e0326aa4e411d69b9309bec668ba90facbcab/src/Serval/src/Serval.Translation/Services/TranslationPlatformServiceV1.cs#L312)):*

Is SF still using the obsolete `Corpus` model? If the `ParallelCorpusId` is actually an obsolete `Corpus` id, then setting `ParallelCorpusRef` to `ParallelCorpusId` would be incorrect. To avoid this, we could check that any of the `ParallelCorpusId` values exists in `Pretranslate.ParallelCorpusRef` in the `UpdateAsync` filter. This would be a way of ensuring that the build is using the newer `ParallelCorpus` model. If it is using the obsolete model, then it simply wouldn't set the analysis.
</blockquote></details>

Done. SF not using the obsolete `Corpus` model. I have updated the logic to check against the engine's list of parallel corpora.

ddaspit (approved):
:lgtm:

> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/718#-:-OXZKulI8YK_0sFTSTuB:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #809, opened 2025-10-24
Update remark for paragraph breaks to match Scripture Forge

Nateowami:
This is what the Scripture Forge UI looks like

localhost_5000_projects_68fa47e15227ca2998b1b953_draft-generation (1)

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/809)

### DIFF
--- a/src/Serval/src/Serval.Translation/Services/PretranslationService.cs
+++ b/src/Serval/src/Serval.Translation/Services/PretranslationService.cs
@@ -351,13 +351,13 @@ public class PretranslationService(
     /// <para>Remarks are generated in the format:</para>
     /// <list type="bullet">
     /// <item><description>
-    /// Paragraph, embed and style markers were moved to the end of the verse.
+    /// Paragraph breaks, embed markers, and style markers were moved to the end of the verse.
     /// </description></item>
     /// <item><description>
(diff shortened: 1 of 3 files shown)


Enkidu93:
> @Enkidu93 What do you envision as a longer-term solution?
SF should supply the remarks. We don't want Serval to be responsible for user-facing text. Also, it's appropriate for Serval to use more technically correct language in places where SF may want to use other language that is more interpretable by end users - I'm thinking here more of the other remark-related PR you put in.

> The decision to call this "Paragraph breaks" was made in the IDQ meeting, and it was mentioned that Serval should follow the same terminology.
OK! I remember the switch to 'paragraph breaks' in the SF UI in a meeting. Was updating the language in the Serval remarks a decision from the IDQ meeting I missed while I was on vacation? Like I said, I'm fine with this as long as we're planning a better long-term solution and this just temporary. If this isn't urgent, we can just wait for Damien's approval.

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/809#-:-OdFm33h0fTdJvClOpYG:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Issue #813, opened 2025-10-30
Change default `embedBehavior` to `Strip`


Enkidu93:
Since we are not currently pretranslating footnotes and bringing embeds and embed text through from the source is probably not the behavior most clients would want (including SF), we should make the default `Strip`.

PR #818, opened 2025-11-05
Strip embeds by default

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/813

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/818)


Enkidu93:
> This is a breaking change. Will it affect Scripture Forge?
> 
> > @ddaspit reviewed 2 of 2 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/serval/818#-:-OdOaa4s9b_ndSd611vS:b7vuz6g)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

SF is changing their behavior to set `embedBehavior` to `Strip`, so as long as their update goes in before ours, this won't change SF's behavior directly. My understanding from the IDQ meeting was that this has been the intended behavior since the initial IDQ marker write-up but that we overlooked making the change. The EITL team was surprised to find embeds being carried over and we decided that it would make sense for us both to change the default and for SF to explicitly set the `embedBehavior`.

PR #835, opened 2025-12-04
Queue time reporting

### LINKED ISSUE #805: Visibility into queue time
Currently, there is no clear way to see how long a build was queued through the API. Since builds only have a `dateFinished` but not a `dateStarted` \ `dateQueued` \ `dateCreated`, it's difficult to retrospectively calculate this. Nathaniel requested this for greater visibility. We could add another timestamp or add another phase for the build to track when it's queued.

pmachapman:
Fixes #805

I chose to put the date started and date completed in the model as properties, rather than execution data, as these values come from ClearML's API directly, and are strongly typed through the various layers. If you think that is not a good idea, let me know and I can change it.

This PR requires https://github.com/sillsdev/machine.py/pull/249

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/835)

### DIFF
--- a/src/Machine/src/Serval.Machine.Shared/Models/ClearMLTask.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Models/ClearMLTask.cs
@@ -23,6 +23,8 @@ public record ClearMLTask
     public string? StatusReason { get; init; }
     public string? StatusMessage { get; init; }
     public required DateTime Created { get; init; }
+    public DateTime? Started { get; init; }
+    public DateTime? Completed { get; init; }
     public int? LastIteration { get; init; }
     public int ActiveDuration { get; init; }
     public required IReadOnlyDictionary<
--- a/src/Machine/src/Serval.Machine.Shared/Services/ClearMLMonitorService.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/ClearMLMonitorService.cs
@@ -1,4 +1,4 @@
-namespace Serval.Machine.Shared.Services;
+﻿namespace Serval.Machine.Shared.Services;
 
 public class ClearMLMonitorService(
     IServiceProvider services,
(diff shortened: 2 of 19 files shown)


Enkidu93 (changes requested):
Thank you, Peter!

> @Enkidu93 reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/835#-:-OfeKOWWD8FhnTMEcs-z:b-i3f2wj)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit and @pmachapman)

___
*[`src/Serval/src/Serval.Translation/Models/Build.cs` line 22 at r1](https://reviewable.io/reviews/sillsdev/serval/835#-OfeJgcKDI2B5VqekFZ6:-OfeJgcKDI2B5VqekFZ7:b-hosqgq) ([raw file](https://github.com/sillsdev/serval/blob/f5e61a5d63b655cc97636a5e1c137d9ddad220e0/src/Serval/src/Serval.Translation/Models/Build.cs#L22)):*
> ```Smalltalk
>     public ExecutionData ExecutionData { get; init; } = new ExecutionData();
>     public DateTime? DateCreated { get; set; }
>     public DateTime? DateStarted { get; set; }
> ```

Is there a way to capture this information in `Phases` rather than separately? I don't mind it being in the model, but it'd be nice to have all the info related to build progress in one place.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/835)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`src/Serval/src/Serval.Shared/Models/BuildPhase.cs` line 14 at r1](https://reviewable.io/reviews/sillsdev/serval/835#-OfjhvWb26IsSdpkQsPA:-Ofuvm134FyB8oW1664U:b-tlp8k1) ([raw file](https://github.com/sillsdev/serval/blob/f5e61a5d63b655cc97636a5e1c137d9ddad220e0/src/Serval/src/Serval.Shared/Models/BuildPhase.cs#L14)):*

I'm guessing that you didn't put a completed timestamp for a phase, because it is redundant. Is that correct?
</blockquote></details>

There wasn't much difference between the two phases complete/start, and between the final phase completing and the task completing (around 10-15 seconds, give or take) in my tests. I thought the time divisions of task.datecreated, task.datestarted, phase1.datestarted, phase2.datestarted, and task.datecompleted was enough details to see where any  delays were taking place?

PR #853, opened 2026-01-14
Set the language codes in the build execution data for echo builds

### LINKED ISSUE #840: Record source and target language in the build
As the source and target language can change for a translation engine, we should record the target and source language in the build so we can later see what these were for the build when it was run.

If possible also record the script.

pmachapman:
Fixes #840

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/853)

### DIFF
--- a/src/Echo/src/EchoEngine/TranslationEngineServiceV1.cs
+++ b/src/Echo/src/EchoEngine/TranslationEngineServiceV1.cs
@@ -121,11 +121,19 @@ public class TranslationEngineServiceV1(
                         cancellationToken: linkedCts.Token
                     );
 
+                    int trainCount = 0;
+                    int pretranslateCount = 0;
+
                     List<InsertPretranslationsRequest> pretranslationsRequests = [];
                     await _parallelCorpusPreprocessingService.PreprocessAsync(
                         request.Corpora.Select(Map).ToList(),
-                        (_, _) => Task.CompletedTask,
-                        (row, _, corpus) =>
+                        (row, _) =>
+                        {
+                            if (row.SourceSegment.Length > 0 && row.TargetSegment.Length > 0)
+                                trainCount++;
+                            return Task.CompletedTask;
+                        },
+                        (row, isInTrainingData, corpus) =>
                         {
                             string[] tokens = row.SourceSegment.Split();
                             pretranslationsRequests.Add(
@@ -146,6 +154,8 @@ public class TranslationEngineServiceV1(
                                     }
                                 }
                             );
+                            if (row.SourceSegment.Length > 0 && !isInTrainingData)
+                                pretranslateCount++;
                             if (cts.IsCancellationRequested)
                                 throw new OperationCanceledException(cts.Token);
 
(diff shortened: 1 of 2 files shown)


pmachapman (commented):
> @pmachapman made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/853)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ddaspit and @Enkidu93).

___
*[`src/Echo/src/EchoEngine/TranslationEngineServiceV1.cs` line 135 at r1](https://reviewable.io/reviews/sillsdev/serval/853#-OixZisC6BMOvcJkNYc8:-Oixl78h941_0pQetK84:b-t4y4gm) ([raw file](https://github.com/sillsdev/serval/blob/988d64a0fdf959efd285a485c562c2c94a734f46/src/Echo/src/EchoEngine/TranslationEngineServiceV1.cs#L135)):*

I wonder about making these values reflect the actual number of training rows/inference rows even though the Echo engine isn't really using that data per se. Do you think it makes more sense just to have them be 0? 

If you did the `UpdateBuildExecutionDataAsync` call after the parallel corpus preprocessing, you could get those values during preprocessing like we do with the true MT engine builds and incorporate them in the execution data. 

...but maybe this spirals into including other 'realistic' values like those for the quote convention analysis and build warnings which we probably do not want to do. On the other hand, getting these values would be easy since we're already running preprocessing on the corpora. What do you think?
</blockquote></details>

Done. Sounds good. Thank you!

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/853#-:-Oj2_9pC6QnaNTgfNvH8:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman).

pmachapman (commented):
> @pmachapman made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/853)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ddaspit and @Enkidu93).

___
*[`src/Echo/src/EchoEngine/TranslationEngineServiceV1.cs` line 135 at r1](https://reviewable.io/reviews/sillsdev/serval/853#-OixZisC6BMOvcJkNYc8:-OjHdzEaAKaJWKXxJqN-:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/988d64a0fdf959efd285a485c562c2c94a734f46/src/Echo/src/EchoEngine/TranslationEngineServiceV1.cs#L135)):*

Thank you! I think we should mirror the counting of the training rows and inference rows here exactly as in https://github.com/sillsdev/serval/blob/226d952c7ad3e1055f9832c86f8a7e7747a0d61e/src/Machine/src/Serval.Machine.Shared/Services/TranslationPreprocessBuildJob.cs#L54 and https://github.com/sillsdev/serval/blob/226d952c7ad3e1055f9832c86f8a7e7747a0d61e/src/Machine/src/Serval.Machine.Shared/Services/WordAlignmentPreprocessBuildJob.cs#L54 - notice that we don't increment on every call.
</blockquote></details>

Done.

PR #862, opened 2026-01-29
Add outbox retry timeout

### LINKED ISSUE #834: Outbox is incorrectly retrying to process messages
We saw this error on production recently:
```
fail: SIL.ServiceToolkit.Services.OutboxDeliveryService[0]
      Attempt 513772.  Failed to process message ID: CancelBuild with content {"engineType":"Nmt","engineId":ID} and error message: Status(StatusCode="Aborted", Detail="There is no build currently running.")
      Grpc.Core.RpcException: Status(StatusCode="Aborted", Detail="There is no build currently running.")
         at Serval.Translation.Consumers.EngineCancelBuildConsumer.HandleMessageAsync(CancelBuildRequest content, Stream stream, CancellationToken cancellationToken) in /app/src/Serval/src/Serval.Translation/Consumers/EngineCancelBuildConsumer.cs:line 18
         at SIL.ServiceToolkit.Services.OutboxDeliveryService.ProcessGroupMessagesAsync(Dictionary`2 consumers, IRepository`1 messages, OutboxMessage message, CancellationToken cancellationToken) in /app/src/ServiceToolkit/src/SIL.ServiceToolkit/Services/OutboxDeliveryService.cs:line 128
         at SIL.ServiceToolkit.Services.OutboxDeliveryService.ProcessMessagesAsync(Dictionary`2 consumers, IRepository`1 messages, CancellationToken cancellationToken) in /app/src/ServiceToolkit/src/SIL.ServiceToolkit/Services/OutboxDeliveryService.cs:line 74
info: Grpc.Net.Client.Internal.GrpcCall[3]
      Call failed with gRPC error status. Status code: 'Aborted', Message: 'There is no build currently running.'.
```

Per Damien:
This should return FAILED_PRECONDITION not ABORTED and we should not retry when ABORTED is returned. 

See thread here: https://sil-language-software.slack.com/archives/C08D9RWTY6N/p1764596491646579

pmachapman:
Fixes #834

To test locally, stop the `serval-machine-engine` container, and start a build (for example using the API example). Messages will then retry with an exponential back off.

Restarting `serval-machine-engine` will cause the messages succeed on retry when the timeout is hit. Previously, messages would not retry in a service outage like this, blocking the queue.

Due to the way the subscription appears to work (watching the first message in the queue for update or deletion), the timeout is in effect as long as there are messages in the queue, and the sending of the first message in the queue has failed. This works correctly now that the attempts counter for that item is no longer incremented (which triggered the subscription, which ran process messages, which failed, which triggered the subscription... ad infinitum).

When the queue is empty, the arrival of the first message in the queue will commence processing, as per before.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/862)

### DIFF
--- a/src/ServiceToolkit/src/SIL.ServiceToolkit/Models/OutboxMessage.cs
+++ b/src/ServiceToolkit/src/SIL.ServiceToolkit/Models/OutboxMessage.cs
@@ -11,5 +11,4 @@ public record OutboxMessage : IEntity
     public required string Content { get; init; }
     public required bool HasContentStream { get; init; }
     public DateTimeOffset Created { get; init; } = DateTimeOffset.UtcNow;
-    public int Attempts { get; init; }
 }
--- a/src/ServiceToolkit/test/SIL.ServiceToolkit.Tests/Services/OutboxDeliveryServiceTests.cs
+++ b/src/ServiceToolkit/test/SIL.ServiceToolkit.Tests/Services/OutboxDeliveryServiceTests.cs
@@ -31,10 +31,6 @@ public class OutboxDeliveryServiceTests
         // Timeout is long enough where the message attempt will be incremented, but not deleted.
         EnableConsumerFailure(env.Consumer2, StatusCode.Internal);
         await env.ProcessMessagesAsync();
-        // Each group should try to send one message
-        Assert.That(env.Messages.Get("B").Attempts, Is.EqualTo(0));
-        Assert.That(env.Messages.Get("A").Attempts, Is.EqualTo(0));
-        Assert.That(env.Messages.Get("C").Attempts, Is.EqualTo(0));
 
         // with now shorter timeout, the messages will be deleted.
         // 4 start build attempts, and only one build completed attempt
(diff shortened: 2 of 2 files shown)


Enkidu93 (approved):
> The other components that use a `MongoSubscription`, such as `BuildService.GetNewerRevisionAsync()` depend on watching for updates to a document, so I figured it was best to not modify `MongoSubscription`, and just stop the `OutboxDeliveryService` from modifying the document. It also makes conceptual sense to have a subscription fire when the objects are updated, I think, as the collection of objects in memory should match the objects in MongoDB, including any updates or deletions.

Ok, I think I was trying to ask a different question than you are answering haha. I was looking at the code in the `MongoSubscription` and it should be listening for inserts if there are no documents when the subscription is started. I just wanted to make sure that your initial comment didn't mean that the subscription wasn't working properly or was getting stuck in situations where it should be catching inserts. But re-reading now, I don't think that's what you were saying, so nevermind :sweat_smile:  - thank you for bearing with me!

> Only when the status code is `Unavailable`, `Unauthenticated`, `PermissionDenied`, or `Cancelled` (these result from some form of service outage). ...
Right, OK, that makes sense. 

> I started down that path, but due the complexity of storing per-message group values and re-triggering processing from these values, I thought it was better to leave that change for a later ...
I see - that makes sense. That would have been hairy to implement for sure! And likely, there aren't many cases besides the cases of a general service outage where we would want to retry anyways. 

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/862#-:-OkTn-x8BFcdxzd7oOJI:b-njjshe)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit).

PR #876, opened 2026-02-18
Downgrade Hangfire to 1.8.22 and Hangfire.Mongo to 1.12.2

pmachapman:
In my testing, it appears that the `OperationCanceledException` that are cancelling jobs are coming from Hangfire directly, when reading the JobState.

It looks like a bug in the implementation for the new `StateHistory` collection in Hangfire.Mongo reading the incorrect state (perhaps because it is reading from a shard in the MongoDB Atlas cluster that does not yet have the latest StateHistory document for the jobId?), so I have downgraded to the version of Hangfire and Hangfire.Mongo before this change was made.

In my testing, this stopped the jobs being cancelled.

The main downside of this PR is that the `serval_jobs` and `machine_jobs` collections will need to be dropped on internal QA and external QA when this PR is deployed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/876)

### DIFF
--- a/src/Machine/src/Serval.Machine.Shared/Serval.Machine.Shared.csproj
+++ b/src/Machine/src/Serval.Machine.Shared/Serval.Machine.Shared.csproj
@@ -34,8 +34,8 @@
 		<PackageReference Include="CommunityToolkit.HighPerformance" Version="8.4.0" />
 		<PackageReference Include="Grpc.AspNetCore" Version="2.76.0" />
 		<PackageReference Include="Grpc.AspNetCore.HealthChecks" Version="2.76.0" />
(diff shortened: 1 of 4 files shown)


ddaspit (approved):
If you have enough information, it would be good to submit an issue to the repo. I'm guessing you tried version 1.13.1. It looks like it had a fix for some issues related state history. 

> @ddaspit reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/876#-:-OllA9yj1JnWFuLtMKQp:bglnd6m)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on pmachapman).

PR #900, opened 2026-04-08
Configure nswag to specify UTC when serializing DateTime

### LINKED ISSUE #899: `created-after` query parameter in the `GetAllBuildsCreatedAfterAsync` endpoint does not preserve time zone
`Serval.Client` does not properly preserve the time zone of the `created-after` parameter. The `DateTime.Kind` property is set to `Unspecified`.

pmachapman:
Fixes #899

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/900)

### DIFF
--- a/src/Serval/src/Serval.ApiServer/nswag.json
+++ b/src/Serval/src/Serval.ApiServer/nswag.json
@@ -46,7 +46,7 @@
       "generateContractsOutput": false,
       "contractsNamespace": null,
       "contractsOutputFilePath": null,
-      "parameterDateTimeFormat": "s",
+      "parameterDateTimeFormat": "u",
       "parameterDateFormat": "yyyy-MM-dd",
       "generateUpdateJsonSerializerSettingsMethod": true,
       "useRequestAndResponseSerializationSettings": false,
--- a/src/Serval/src/Serval.Client/Client.g.cs
+++ b/src/Serval/src/Serval.Client/Client.g.cs
@@ -2278,7 +2278,7 @@ namespace Serval.Client
                     urlBuilder_.Append('?');
                     if (createdAfter != null)
                     {
-                        urlBuilder_.Append(System.Uri.EscapeDataString("created-after")).Append('=').Append(System.Uri.EscapeDataString(createdAfter.Value.ToString("s", System.Globalization.CultureInfo.InvariantCulture))).Append('&');
+                        urlBuilder_.Append(System.Uri.EscapeDataString("created-after")).Append('=').Append(System.Uri.EscapeDataString(createdAfter.Value.ToString("u", System.Globalization.CultureInfo.InvariantCulture))).Append('&');
                     }
                     urlBuilder_.Length--;
 
@@ -2380,7 +2380,7 @@ namespace Serval.Client
                     urlBuilder_.Append('?');
                     if (finishedAfter != null)
                     {
-                        urlBuilder_.Append(System.Uri.EscapeDataString("finished-after")).Append('=').Append(System.Uri.EscapeDataString(finishedAfter.Value.ToString("s", System.Globalization.CultureInfo.InvariantCulture))).Append('&');
+                        urlBuilder_.Append(System.Uri.EscapeDataString("finished-after")).Append('=').Append(System.Uri.EscapeDataString(finishedAfter.Value.ToString("u", System.Globalization.CultureInfo.InvariantCulture))).Append('&');
                     }
                     urlBuilder_.Length--;
 


ddaspit (changes requested):
> @ddaspit reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/900)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93 and pmachapman).

___
*[`src/Serval/src/Serval.ApiServer/nswag.json` line 49 at r1](https://reviewable.io/reviews/sillsdev/serval/900#-OpmZ9Ao5eDruRPIvpva:-OpmZ9Ao5eDruRPIvpvb:bw9a8hz) ([raw file](https://github.com/sillsdev/serval/blob/efc4f18cfe2269cd660d189ac611f1b3722721a6/src/Serval/src/Serval.ApiServer/nswag.json#L49)):*
> ```JSON
>       "contractsNamespace": null,
>       "contractsOutputFilePath": null,
>       "parameterDateTimeFormat": "u",
> ```

My concern with this approach is that it will only work properly if the `DateTime` is already in [UTC](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#the-universal-sortable-u-format-specifier) time. To make this robust, we would need to convert the `DateTime` to UTC before serializing using the `u` format. Is there any way to perform the conversion in `Serval.Client`? `DateTimeOffset` does perform the conversion automatically, so maybe we should use that instead.

pmachapman (commented):
> @pmachapman made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/900)* status: all files reviewed, 1 unresolved discussion (waiting on ddaspit and Enkidu93).

___
*[`src/Serval/src/Serval.ApiServer/nswag.json` line 49 at r1](https://reviewable.io/reviews/sillsdev/serval/900#-OpmZ9Ao5eDruRPIvpva:-Oq2nK-rAd6hcVucnwBw:b-bstbz) ([raw file](https://github.com/sillsdev/serval/blob/efc4f18cfe2269cd660d189ac611f1b3722721a6/src/Serval/src/Serval.ApiServer/nswag.json#L49)):*

My concern with this approach is that it will only work properly if the `DateTime` is already in [UTC](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#the-universal-sortable-u-format-specifier) time. To make this robust, we would need to convert the `DateTime` to UTC before serializing using the `u` format. Is there any way to perform the conversion in `Serval.Client`? `DateTimeOffset` does perform the conversion automatically, so maybe we should use that instead.
</blockquote></details>

Based on the configuration below, `"dateType": "System.DateTimeOffset",`, in `Client.g.cs`, a `DateTimeOffset` is passed to `GetAllBuildsCreatedAfterAsync` and `GetNextFinishedBuildAsync`, so `.ToString("u")` will automatically convert from whatever timezone is specified to UTC, so from what I understand it is already converting?

If you want an explicit conversion in `BuildService`, I can update `IBuildService` to accept `DateTimeOffset`, which won't change `Client.g.cs`, as it already accepts `DateTimeOffset`.

ddaspit (approved):
:lgtm:

> @ddaspit reviewed all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/900#-:-Oq6NSlA2XDm2w_q1KVr:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Enkidu93).

___
*[`src/Serval/src/Serval.ApiServer/nswag.json` line 49 at r1](https://reviewable.io/reviews/sillsdev/serval/900#-OpmZ9Ao5eDruRPIvpva:-Oq6N9qb8535heTzoRHO:b-npna4h) ([raw file](https://github.com/sillsdev/serval/blob/efc4f18cfe2269cd660d189ac611f1b3722721a6/src/Serval/src/Serval.ApiServer/nswag.json#L49)):*

Based on the configuration below, `"dateType": "System.DateTimeOffset",`, in `Client.g.cs`, a `DateTimeOffset` is passed to `GetAllBuildsCreatedAfterAsync` and `GetNextFinishedBuildAsync`, so `.ToString("u")` will automatically convert from whatever timezone is specified to UTC, so from what I understand it is already converting?

If you want an explicit conversion in `BuildService`, I can update `IBuildService` to accept `DateTimeOffset`, which won't change `Client.g.cs`, as it already accepts `DateTimeOffset`.
</blockquote></details>

You're right. I missed that `createdAfter` and `finishedAfter` are `DateTimeOffset` instances and not `DateTime` instances.

Enkidu93 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/900#-:-Oq6iCur9-3sdOGg4hHM:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on pmachapman).

PR #954, opened 2026-06-02
Use one cursor per subscription

Enkidu93:
This addresses a problem we've been receiving alerts about from Atlas where thousands of documents were being scanned but only a few returned. I think this is a problem we've actually hit in the past as well without realizing that this was the source of the problem. 

As I mentioned in the thread on Slack, because we use the `StartAfterOperationTime` based on the cluster time of the last change event (which is only precise down to the second), we can sometimes get stuck in a loop where we keep reprocessing the same update. This will not happen if we keep a single long-running cursor (or if we used a resume token). If the collection goes a long time without an update and we're stuck on an update like this, we end up having to scan the oplog over and over scanning days of changes because the timestamp that governs the `StartAfterOperationTime` never gets updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/954)

### DIFF
--- a/src/DataAccess/src/SIL.DataAccess.Abstractions/IRepository.cs
+++ b/src/DataAccess/src/SIL.DataAccess.Abstractions/IRepository.cs
@@ -28,6 +28,7 @@ public interface IRepository<T>
     Task<ISubscription<T>> SubscribeAsync(
         Expression<Func<T, bool>> filter,
         IEnumerable<(Expression<Func<T, object?>> Field, SortOrder SortOrder)>? sort = null,
+        IReadOnlySet<EntityChangeType>? changeTypes = null,
         CancellationToken cancellationToken = default
     );
 }
--- a/src/DataAccess/src/SIL.DataAccess.Abstractions/IRepository.cs
+++ b/src/DataAccess/src/SIL.DataAccess.Abstractions/IRepository.cs
@@ -28,7 +28,7 @@ public interface IRepository<T>
     Task<ISubscription<T>> SubscribeAsync(
         Expression<Func<T, bool>> filter,
         IEnumerable<(Expression<Func<T, object?>> Field, SortOrder SortOrder)>? sort = null,
-        IReadOnlySet<EntityChangeType>? changeTypes = null,
+        SubscriptionMode mode = SubscriptionMode.Entity,
         CancellationToken cancellationToken = default
     );
 }
--- a/src/DataAccess/src/SIL.DataAccess.Abstractions/ISubscription.cs
+++ b/src/DataAccess/src/SIL.DataAccess.Abstractions/ISubscription.cs
@@ -4,9 +4,6 @@ public interface ISubscription<T> : IDisposable
     where T : IEntity
 {
     EntityChange<T> Change { get; }
-    Task WaitForChangeAsync(
-        TimeSpan? timeout = null,
-        IReadOnlySet<EntityChangeType>? changeTypes = null,
-        CancellationToken cancellationToken = default
-    );
+    Task WaitForChangeAsync(TimeSpan? timeout = null, CancellationToken cancellationToken = default);
+    DateTime Timestamp { get; }
 }
--- a/src/DataAccess/src/SIL.DataAccess.Abstractions/ISubscription.cs
+++ b/src/DataAccess/src/SIL.DataAccess.Abstractions/ISubscription.cs
@@ -5,5 +5,4 @@ public interface ISubscription<T> : IDisposable
 {
     EntityChange<T> Change { get; }
     Task WaitForChangeAsync(TimeSpan? timeout = null, CancellationToken cancellationToken = default);
-    DateTime Timestamp { get; }
 }
--- a/src/DataAccess/src/SIL.DataAccess.Abstractions/SubscriptionMode.cs
+++ b/src/DataAccess/src/SIL.DataAccess.Abstractions/SubscriptionMode.cs
@@ -1,0 +1,8 @@
+namespace SIL.DataAccess;
+
+public enum SubscriptionMode
+{
+    Entity,
+
+    Repository,
+}
--- a/src/DataAccess/src/SIL.DataAccess/MemoryRepository.cs
+++ b/src/DataAccess/src/SIL.DataAccess/MemoryRepository.cs
@@ -299,7 +299,7 @@ public class MemoryRepository<T> : IRepository<T>
     public async Task<ISubscription<T>> SubscribeAsync(
         Expression<Func<T, bool>> filter,
         IEnumerable<(Expression<Func<T, object?>> Field, SortOrder SortOrder)>? sort = null,
-        IReadOnlySet<EntityChangeType>? changeTypes = null,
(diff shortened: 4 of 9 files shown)


pmachapman (approved):
:lgtm: I ran it with my PR for SF that uses GetNextFinishedBuild, and it ran well for the scenarios I tested (starting builds, finishing builds, deleting builds).

> @pmachapman reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/954#-:-OuABlkUAzdZD5Cw0oUA:bmvso61)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ddaspit).

ddaspit (commented):
> @ddaspit reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/954)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93).

___
*[`src/DataAccess/src/SIL.DataAccess.Abstractions/IRepository.cs` line 31 at r1](https://reviewable.io/reviews/sillsdev/serval/954#-OuDSLkbAvDLCh8VzWdb:-OuDSLkbAvDLCh8VzWdc:bzb53ga) ([raw file](https://github.com/sillsdev/serval/blob/d6c8538d0fbd6287a50e0dc426de7c03fb5e21a3/src/DataAccess/src/SIL.DataAccess.Abstractions/IRepository.cs#L31)):*
> ```Smalltalk
>         Expression<Func<T, bool>> filter,
>         IEnumerable<(Expression<Func<T, object?>> Field, SortOrder SortOrder)>? sort = null,
>         IReadOnlySet<EntityChangeType>? changeTypes = null,
> ```

When `changeTypes` is specified, the subscription works very differently. It changes the subscription to watch for changes on a whole repository. This is useful implementing queueing or getting the next finished build. It is unclear from the API that the subscription is running in a different mode. The original implementation was designed to watch for changes on a single entity. This worked properly with timestamps. In any case, I think it would be good to change this parameter to an enum type called `mode`. The enum would have two values: `Entity` and `Repository`. `Entity` would be the default.

Enkidu93 (commented):
I think I've incorporated the feedback you guys had in Slack. Let me know how this looks!

> @Enkidu93 made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/954#-:-OuDl6QjDSJYnCtf1G3Z:b-e3y3il)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on ddaspit and pmachapman).

___
*[`src/DataAccess/src/SIL.DataAccess.Abstractions/IRepository.cs` line 31 at r1](https://reviewable.io/reviews/sillsdev/serval/954#-OuDSLkbAvDLCh8VzWdb:-OuDkzPg7PSG5HzYnWfU:b-63y3tc) ([raw file](https://github.com/sillsdev/serval/blob/d6c8538d0fbd6287a50e0dc426de7c03fb5e21a3/src/DataAccess/src/SIL.DataAccess.Abstractions/IRepository.cs#L31)):*

When `changeTypes` is specified, the subscription works very differently. It changes the subscription to watch for changes on a whole repository. This is useful implementing queueing or getting the next finished build. It is unclear from the API that the subscription is running in a different mode. The original implementation was designed to watch for changes on a single entity. This worked properly with timestamps. In any case, I think it would be good to change this parameter to an enum type called `mode`. The enum would have two values: `Entity` and `Repository`. `Entity` would be the default.
</blockquote></details>

Done. Let me know if this is what you were thinking.

pmachapman (approved):
:lgtm: I reran with the same tests as previously, and it (still) runs great. I still can't remember at all why resume tokens didn't work for me :-/

> @pmachapman reviewed 9 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/954#-:-OuFIkID90OylOAoWxW-:b-ts2upy)* status: all files reviewed, 1 unresolved discussion (waiting on ddaspit).

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 9 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/954#-:-OuHmu3O7K8hCjcOZ2kC:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Enkidu93).

___
*[`src/DataAccess/src/SIL.DataAccess/MongoSubscription.cs` line 147 at r2](https://reviewable.io/reviews/sillsdev/serval/954#-OuHmSdtBRHwr0Gt1OeG:-OuHmSdtBRHwr0Gt1OeH:b-tzjj0m) ([raw file](https://github.com/sillsdev/serval/blob/823f1c007efd99f0d5cc28eadd8b9bcad4cf4fa7/src/DataAccess/src/SIL.DataAccess/MongoSubscription.cs#L147)):*
> ```Smalltalk
>             }
>         }
>         catch (MongoException)
> ```

Nit: The cursor gets disposed when the subscription is disposed, so this catch isn't really necessary.

Enkidu93:
*[`src/DataAccess/src/SIL.DataAccess/MongoSubscription.cs` line 147 at r2](https://reviewable.io/reviews/sillsdev/serval/954#-OuHmSdtBRHwr0Gt1OeG:-OuHnsfw3r43rIY_qbRi:b6dn6xl) ([raw file](https://github.com/sillsdev/serval/blob/823f1c007efd99f0d5cc28eadd8b9bcad4cf4fa7/src/DataAccess/src/SIL.DataAccess/MongoSubscription.cs#L147)):*

Nit: The cursor gets disposed when the subscription is disposed, so this catch isn't really necessary.
</blockquote></details>

@ddaspit, I was thinking. What if you hit an exception on a particular `WaitForChangeAsync()` but you catch it in the calling code and then call `WaitForChangeAsync()` again. You'll want the old cursor to be disposed and `_cursor` to be `null` so that you will recreate it and use the resume token when you call this method again, right?

ddaspit (commented):
> @ddaspit made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/954)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Enkidu93).

___
*[`src/DataAccess/src/SIL.DataAccess/MongoSubscription.cs` line 147 at r2](https://reviewable.io/reviews/sillsdev/serval/954#-OuHmSdtBRHwr0Gt1OeG:-OuHpsqTAnusGrFKAVsF:bct9p11) ([raw file](https://github.com/sillsdev/serval/blob/823f1c007efd99f0d5cc28eadd8b9bcad4cf4fa7/src/DataAccess/src/SIL.DataAccess/MongoSubscription.cs#L147)):*

@ddaspit, I was thinking. What if you hit an exception on a particular `WaitForChangeAsync()` but you catch it in the calling code and then call `WaitForChangeAsync()` again. You'll want the old cursor to be disposed and `_cursor` to be `null` so that you will recreate it and use the resume token when you call this method again, right?
</blockquote></details>

That's a good point. In any case, there is no drawback to leaving it.

PR #980, opened 2026-06-25
Block invalid language tags

### LINKED ISSUE #967: RCE via `SourceLanguage` / `TargetLanguage`
**File:** `src/Machine/src/Serval.Machine.Shared/Services/NmtClearMLBuildJobFactory.cs:33-41`

Same sink, different source:

```csharp
_languageTagService.ConvertToFlores200Code(engine.SourceLanguage, out string srcLang);
_languageTagService.ConvertToFlores200Code(engine.TargetLanguage, out string trgLang);
...
+ $"    'src_lang': '{srcLang}',\n"
+ $"    'trg_lang': '{trgLang}',\n"
```

`ConvertToFlores200Code` falls back to the **raw, unmodified input** when parsing fails (`src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs:44-45`), and `SourceLanguage`/`TargetLanguage` are `required string` with **no validation** at engine creation (`src/Serval/src/Serval.Translation/Features/Engines/CreateEngine.cs`). An input such as `en'; import os; os.system(...); x='` fails the strict tag parser (so the raw value is returned) and breaks out of the single-quoted Python literal → RCE.

**Shared root cause / fix:** C1 and C2 share one root cause — **building executable Python by string interpolation of externally-controlled data**. The durable fix is to pass all dynamic values to the Python job as data (ClearML hyperparameters or a single encoded JSON blob the Python side decodes), combined with strict input validation: validate `SourceLanguage`/`TargetLanguage` against a strict BCP-47 pattern (e.g. `^[A-Za-z0-9-]{1,35}$`) and reject non-conforming tags rather than echoing raw input. These are the top priority.

pmachapman:
Fixes #967.

I also took the opportunity to improve unit testing and API documentation around this area.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/980)

### DIFF
--- a/src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs
+++ b/src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs
@@ -1,7 +1,10 @@
 ﻿namespace Serval.Machine.Shared.Services;
 
-public class LanguageTagService : ILanguageTagService
+public partial class LanguageTagService : ILanguageTagService
 {
+    [GeneratedRegex("^[A-Za-z0-9_-]+$")]
+    private static partial Regex BroadlyPermissibleLanguageTagPattern();
+
     private readonly HashSet<string> _flores200Languages = [];
     private readonly HashSet<string> _flores200Scripts = [];
     private readonly LanguageTagParser _parser = new();
@@ -40,9 +43,16 @@ public class LanguageTagService : ILanguageTagService
     public Flores200Support ConvertToFlores200Code(string languageTag, out string flores200Code)
     {
         if (_parser.TryParse(languageTag, out string? languageCode, out string? scriptCode))
+        {
             flores200Code = $"{languageCode}_{scriptCode}";
+        }
         else
+        {
+            if (!BroadlyPermissibleLanguageTagPattern().IsMatch(languageTag))
+                throw new InvalidOperationException($"Invalid language tag: {languageTag}");
             flores200Code = languageTag;
+        }
+
         if (_flores200Scripts.Contains(scriptCode ?? ""))
         {
             if (_flores200Languages.Contains(flores200Code))
--- a/src/Machine/test/Serval.Machine.Shared.Tests/Services/LanguageTagServiceTests.cs
+++ b/src/Machine/test/Serval.Machine.Shared.Tests/Services/LanguageTagServiceTests.cs
@@ -10,7 +10,6 @@ public class LanguageTagServiceTests
             Sldr.Initialize();
     }
 
-    [Test]
     [TestCase("es", "spa_Latn", Description = "Iso639_1Code")]
     [TestCase("hne", "hne_Deva", Description = "Iso639_3Code")]
     [TestCase("ks-Arab", "kas_Arab", Description = "ScriptCode")]
@@ -27,6 +26,7 @@ public class LanguageTagServiceTests
     [TestCase("eng-Latn", "eng_Latn", Description = "DashToUnderscore")]
     [TestCase("kor", "kor_Hang", Description = "KoreanScript")]
     [TestCase("kor_Kore", "kor_Hang", Description = "KoreanScriptCorrection")]
+    [TestCase("unknown", "unknown", Description = "UnknownLanguage")]
     public void ConvertToFlores200CodeTest(string language, string internalCodeTruth)
     {
         new LanguageTagService().ConvertToFlores200Code(language, out string internalCode);
@@ -34,6 +34,11 @@ public class LanguageTagServiceTests
     }
 
     [Test]
+    public void ConvertToFlores200CodeTest_InvalidLanguage() =>
+        Assert.Throws<InvalidOperationException>(() =>
+            new LanguageTagService().ConvertToFlores200Code("'Invalid Language'", out string _)
+        );
+
     [TestCase("en", "eng_Latn", Flores200Support.LanguageAndScript)]
     [TestCase("ms", "zsm_Latn", Flores200Support.LanguageAndScript)]
     [TestCase("cmn", "zho_Hans", Flores200Support.LanguageAndScript)]
--- a/src/Serval/src/Serval.Client/Client.g.cs
+++ b/src/Serval/src/Serval.Client/Client.g.cs
@@ -2186,7 +2186,7 @@ namespace Serval.Client
 
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         /// <summary>
-        /// Get information regarding a language for a given engine type
+        /// Get information regarding a language for a given engine type.
         /// </summary>
         /// <remarks>
         /// This endpoint exists primarily to support `nmt` model-training since `echo` and `smt-transfer` engines support all languages equally. Given a language tag, it provides the ISO 639-3 code that the tag maps to internally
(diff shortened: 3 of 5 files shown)


ddaspit (changes requested):
> @ddaspit reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/980)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93 and pmachapman).

___
*[`src/Serval/src/Serval.Translation/Features/EngineTypes/GetLanguageInfo.cs` line 85 at r1](https://reviewable.io/reviews/sillsdev/serval/980#-Ovz0XQG9RrakgCBDTcq:-Ovz0XQG9RrakgCBDTcr:b-33oaye) ([raw file](https://github.com/sillsdev/serval/blob/dfc4d6ccba83aab89cbe5166b124ed3856b949e2/src/Serval/src/Serval.Translation/Features/EngineTypes/GetLanguageInfo.cs#L85)):*
> ```Smalltalk
>             return NotFound();
>         }
>         catch (ArgumentException)
> ```

This is a very generic exception. I don't want to accidentally mask bugs in the code. Could we use a more specific exception or catch the exception closer to where the `LanguageTagService` is called and indicate that it is invalid in the response?

Enkidu93:
*[`src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs` line 5 at r1](https://reviewable.io/reviews/sillsdev/serval/980#-OvzDoTc-Ti19EBKOrHI:-OvzDoTdDLd1enm6q7KZ:b-ok9fjf) ([raw file](https://github.com/sillsdev/serval/blob/dfc4d6ccba83aab89cbe5166b124ed3856b949e2/src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs#L5)):*
> ```Smalltalk
> public partial class LanguageTagService : ILanguageTagService
> {
>     [GeneratedRegex("^[A-Za-z0-9_-]+$")]
> ```

I wonder how restrictive we want to make this. I'm worried about long codes we get out of Paratext projects settings files like `cod-Scrip-RG-dialect`. We have a number of these engines already in production. I assume if they are a true language code, we won't hit this regex in the code, but if they aren't (maybe a made-up code for a under documented language or new dialect) *and* they use a dialect-differentiator code (not sure what the official name is?) that has characters other than these, we could be preventing them from running a build. Do you know if you're able to only use a-z in these codes? None of the ones on production have something other than a-z. I'm thinking here of the final segment of a `<LanguageIsoCode\>` tag in a Settings.xml. I can't seem to find documentation for this.

Enkidu93 (commented):
> @Enkidu93 reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/980)* status: all files reviewed, 2 unresolved discussions (waiting on pmachapman).

pmachapman (commented):
> @pmachapman made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/980)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on ddaspit and Enkidu93).

___
*[`src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs` line 5 at r1](https://reviewable.io/reviews/sillsdev/serval/980#-OvzDoTc-Ti19EBKOrHI:-OwG3n364-nXrKGJZNzG:bv9kjm3) ([raw file](https://github.com/sillsdev/serval/blob/dfc4d6ccba83aab89cbe5166b124ed3856b949e2/src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs#L5)):*

I wonder how restrictive we want to make this. I'm worried about long codes we get out of Paratext projects settings files like `cod-Scrip-RG-dialect`. We have a number of these engines already in production. I assume if they are a true language code, we won't hit this regex in the code, but if they aren't (maybe a made-up code for a under documented language or new dialect) *and* they use a dialect-differentiator code (not sure what the official name is?) that has characters other than these, we could be preventing them from running a build. Do you know if you're able to only use a-z in these codes? None of the ones on production have something other than a-z. I'm thinking here of the final segment of a `<LanguageIsoCode\>` tag in a Settings.xml. I can't seem to find documentation for this.
</blockquote></details>

I tried to make this broadly support BCP-47 (https://en.wikipedia.org/wiki/IETF_language_tag), with the primary purpose of not allowing injection, rather than ensuring a valid language code (I think the `TryParse` above will take care of known validity). As far as I am aware (based on the Wikipedia article) language codes are alphanumeric with underscores and dashes.
___
*[`src/Serval/src/Serval.Translation/Features/EngineTypes/GetLanguageInfo.cs` line 85 at r1](https://reviewable.io/reviews/sillsdev/serval/980#-Ovz0XQG9RrakgCBDTcq:-OwG5f9rASJ3Vqwt1W7n:b-bzjdnf) ([raw file](https://github.com/sillsdev/serval/blob/dfc4d6ccba83aab89cbe5166b124ed3856b949e2/src/Serval/src/Serval.Translation/Features/EngineTypes/GetLanguageInfo.cs#L85)):*

This is a very generic exception. I don't want to accidentally mask bugs in the code. Could we use a more specific exception or catch the exception closer to where the `LanguageTagService` is called and indicate that it is invalid in the response?
</blockquote></details>

Done. I have changed this to an `InvalidOperationException`, as there is already a `BadRequestExceptionFilter` which will catch this and output a 400 HTTP status code.

Enkidu93:
*[`src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs` line 5 at r1](https://reviewable.io/reviews/sillsdev/serval/980#-OvzDoTc-Ti19EBKOrHI:-OwIb-Ut1IUxj4mrp36X:b-ur304e) ([raw file](https://github.com/sillsdev/serval/blob/dfc4d6ccba83aab89cbe5166b124ed3856b949e2/src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs#L5)):*

I tried to make this broadly support BCP-47 (https://en.wikipedia.org/wiki/IETF_language_tag), with the primary purpose of not allowing injection, rather than ensuring a valid language code (I think the `TryParse` above will take care of known validity). As far as I am aware (based on the Wikipedia article) language codes are alphanumeric with underscores and dashes.
</blockquote></details>

Right, I think it would be an odd case, but I'm just thinking of folks who might supply a yet-unregisterd/made-up language code. Then it would get past the `TryParse`. If it also had a variant specified with non-ASCII letters/numbers, we would throw an error. From the Wikipedia article, it's not clear than the variant names are completely standardized. I guess I was mainly wondering if Paratext itself allows you to enter non-ASCII letters/numbers in these settings because you just pass this tag from the Settings.xml on to Serval from SF, correct?

Enkidu93 (commented):
> @Enkidu93 reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/980)* status: all files reviewed, 2 unresolved discussions (waiting on ddaspit and pmachapman).

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 4 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/980#-:-OwJy1BF6GMmuYAGZ0CX:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93).

___
*[`src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs` line 5 at r1](https://reviewable.io/reviews/sillsdev/serval/980#-OvzDoTc-Ti19EBKOrHI:-OwJqJY8-n71Nv0sGejh:bhuukud) ([raw file](https://github.com/sillsdev/serval/blob/dfc4d6ccba83aab89cbe5166b124ed3856b949e2/src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs#L5)):*

Right, I think it would be an odd case, but I'm just thinking of folks who might supply a yet-unregisterd/made-up language code. Then it would get past the `TryParse`. If it also had a variant specified with non-ASCII letters/numbers, we would throw an error. From the Wikipedia article, it's not clear than the variant names are completely standardized. I guess I was mainly wondering if Paratext itself allows you to enter non-ASCII letters/numbers in these settings because you just pass this tag from the Settings.xml on to Serval from SF, correct?
</blockquote></details>

Paratext does not allow you to enter non-ASCII characters in a private-use subtag. All other subtags must be valid and known.

Enkidu93 (approved):
> @Enkidu93 made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/980)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on pmachapman).

___
*[`src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs` line 5 at r1](https://reviewable.io/reviews/sillsdev/serval/980#-OvzDoTc-Ti19EBKOrHI:-OwJySS_AIEyxYXRfpJb:bosdw10) ([raw file](https://github.com/sillsdev/serval/blob/dfc4d6ccba83aab89cbe5166b124ed3856b949e2/src/Machine/src/Serval.Machine.Shared/Services/LanguageTagService.cs#L5)):*

Paratext does not allow you to enter non-ASCII characters in a private-use subtag. All other subtags must be valid and known.
</blockquote></details>

OK, perfect, thank you!

PR #985, opened 2026-06-25
Add average pretranslation confidence to build execution data

### LINKED ISSUE #937: Reconsider how confidences are stored for an engine
Currently, the confidences are only associated with pretranslations. This means that once a new build is run, the old pretranslations (and thus confidences) are discarded. Seeing how confidences change across time as an individual project drafts new books could be very helpful, but the current behavior would not allow this. Since we probably don't care about tracking per-verse confidences per project across time (since they will likely be drafting new books), we should consider adding some confidence stats (either overall or per book, chapter averages and maybe std dev) to the build itself. This would allow us to study confidences across time without having to keep all the past pretranslations.

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/937

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/985)

### DIFF
--- a/src/Echo/src/EchoEngine/TranslationEngineService.cs
+++ b/src/Echo/src/EchoEngine/TranslationEngineService.cs
@@ -237,6 +237,7 @@ public class TranslationEngineService(BackgroundTaskQueue taskQueue, IParallelCo
 
                     await platform.InsertPretranslationsAsync(
                         engineId,
+                        buildId,
                         pretranslations.ToAsyncEnumerable(),
                         linkedCts.Token
                     );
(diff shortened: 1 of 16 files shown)


ddaspit (approved):
:lgtm:

> @ddaspit reviewed 16 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/985#-:-OwOq2ApDXE6YoZjFBkX:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Enkidu93).

PR #1012, opened 2026-07-28
Structured build diagnostics

### LINKED ISSUE #965: Expose structured build warnings through API
Currently, we provide warnings as a list of string warning messages included in the build execution data. Although these warnings have been helpful, there are a few problems with them:

1. Because we only supply a warning message, SF (or another client) cannot appropriately show these warnings to users (or localize them). SF needs to be responsible for UI text. 
2. The warnings have no indication as to their severity. Some warnings like `"The script for the source language '{code} is not in Flores-200"`, `"Only 100 segments were selected for training."`, or versification warnings that suggest large numbers of verses are missing could have significant impact on the quality of the draft whereas other warnings like a stray extra verse probably won't.
3. The messages do not always provide a clear way for the issue to be rectified or the context necessary to do so: Some warnings may need to be resolved by an admin (e.g. complex versification problems) whereas others may be things that the users can resolve themselves (e.g. 'too little training data - check your training source') and others might be difficult to solve at all (e.g. unknown script). 

Given these, I propose we move to a more structured warning (or maybe 'diagnostic'?) format. For backwards compatibility, we'll retain the `executionData.warnings` property but add a list of `Warning` (or `Diagnostic` or ...?) objects to the `Build` model. Something like:
```
class Warning:
   string id // (or code? e.g. CS0001 kinda thing)
   string message
   enum severity = [info, warn, error] // (or similar - if we use the class name Warning, maybe low, med, high; if Diagnostic, info, warn error)
   Location location:
        string? parallelCorpusId // (could be mapped to URIs/resource links on the front end)
        string? monolingualCorpusId
        string? fileId
        string? filename
        int[]? linenumbers
        string[]? references
```

@Nateowami @ddaspit What do you think?

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/965.

See https://docs.google.com/document/d/1719NBOAx0NRvk8WpcciTqvFPPZDRSK-oZXbrIKUgL20/edit?tab=t.0#heading=h.dsj2hidolq5 for details. 

This work so far does *not* include:
* Adding tests
* Updating documentation - I'm figuring we'll want this in the wiki?
* The correct confidence threshold

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)


ddaspit (changes requested):
> @ddaspit reviewed 31 files and all commit messages, and made 19 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: all files reviewed, 19 unresolved discussions (waiting on Enkidu93).

___
*[a discussion](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeytAJ81ziYsTyMsL_:-OyeytAJ81ziYsTyMsLa:b-aeyg8d) (no related file):*
Are you planning on adding unit tests? And yes, we will want to add a page for the diagnostics to the wiki page. It would probably be good to have a link to the wiki page in the Swagger docs.
___
*[`src/Serval/src/Serval.Translation/Dtos/ExecutionDataDto.cs` line 14 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeyWf8-87jOysiqwBH:-OyeyWf8-87jOysiqwBI:b5s0nq9) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Dtos/ExecutionDataDto.cs#L14)):*
> ```Smalltalk
>     [Obsolete]
>     public IReadOnlyList<string> Warnings { get; init; } = [];
>     public IReadOnlyList<DiagnosticDto> Diagnostics { get; init; } = [];
> ```

This should be optional.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/EchoPreprocessBuildJob.cs` line 59 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeyFmQ2CPfpKMcs9Md:-OyeyFmQ2CPfpKMcs9Me:b-xw2lag) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/EchoPreprocessBuildJob.cs#L59)):*
> ```Smalltalk
>             sourceLanguageTag,
>             targetLanguageTag,
>             true,
> ```

Make the parameter name explicit.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 475 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyeuqgg0CaqSK6xI6Wq:-Oyeuqgg0CaqSK6xI6Wr:b-ai755a) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L475)):*
> ```Smalltalk
>             .Select(b => new Diagnostic
>             {
>                 Code = "MODEL-004",
> ```

I think this is supposed to be `MODEL-003`.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 479 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyey5ke-gDOHXvCWCYp:-Oyey5ke-gDOHXvCWCYq:b11xagr) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L479)):*
> ```Smalltalk
>                 Severity = Models.DiagnosticSeverity.Warn,
>                 Message =
>                     $"The average pretranslation model confidence {b.averageConfidence} in book {b.bookId} is unusually low for the base model {ModelName}",
> ```

This confidence should be formatted.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 488 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyexNH537JvPXlVqdHF:-OyexNH537JvPXlVqdHG:bsxjlxh) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L488)):*
> ```Smalltalk
>             });
> 
>         if (badBookConfidences.Any())
> ```

This deferred LINQ query is getting run multiple times.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 492 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyexu7lEjeBlB6whOUb:-Oyexu7m7hvGWUDUYE61:b-d21n71) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L492)):*
> ```Smalltalk
>             Build? currentBuild = await _builds.GetAsync(b => b.Id == buildId, cancellationToken);
> 
>             await _builds.UpdateAsync(
> ```

Can we perform this update and the next in a single Mongo call?
___
*[`src/Machine/src/Serval.Machine.Shared/Services/EchoWordAlignmentPreprocessBuildJob.cs` line 119 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeyKqm6nIOpIgMBaOa:-OyeyKqn7VYJ2jHLi6co:b-xw2lag) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/EchoWordAlignmentPreprocessBuildJob.cs#L119)):*
> ```Smalltalk
>             sourceLanguageTag,
>             targetLanguageTag,
>             true,
> ```

Make the parameter name explicit.
___
*[`src/Serval/src/Serval.Translation/Dtos/DiagnosticDto.cs` line 3 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyezKLa8lMam6ClM92a:-OyezKLa8lMam6ClM92b:btgkx1w) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Dtos/DiagnosticDto.cs#L3)):*
> ```Smalltalk
> namespace Serval.Translation.Dtos;
> 
> public record DiagnosticDto
> ```

We should call this `TranslationDiagnosticDto`, so the name doesn't collide with the word alignment diagnostic in `Serval.Client`. `ExecutionDataDto` should really be renamed too, but that would be a breaking change at this point.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 203 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevFYvA9U8UOSbV3gc:-OyevFYvA9U8UOSbV3gd:b2l5c50) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L203)):*
> ```Smalltalk
>     }
> 
>     protected virtual IReadOnlyList<BuildDiagnostic> GetDiagnostics(
> ```

We have a `MaxWarnings`. We should probably have a `MaxDiagnostics`.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 225 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevWuEA_DKevPhvv4g:-OyevWuEA_DKevPhvv4h:bqi3hts) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L225)):*
> ```Smalltalk
>                 string versificationName,
>                 IReadOnlyList<UsfmVersificationDiagnosticContract> usfmDiagnostics
>             ) in ParallelCorpusService.AnalyzeUsfmVersification(parallelCorpora)
> ```

I don't love that we are running `AnalyzeUsfmVersification` twice for a build. Could we do this once and pass it to `GetWarnings` and `GetDiagnostics` or derive the warnings from the diagnostics?
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 243 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyev57EBtxN-xVjeh55:-Oyev57EBtxN-xVjeh56:b-j1sbd3) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L243)):*
> ```Smalltalk
>                         Serval.Shared.Contracts.UsfmVersificationDiagnosticType.InvalidChapter => new BuildDiagnostic
>                         {
>                             Code = "USFM-001",
> ```

This is probably overkill, but I would feel a bit less anxious if the number part of the code was 4 digits.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 406 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyetZcA6M1ar9SzQaQe:-OyetZcA6M1ar9SzQaQf:b-wqhlvv) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L406)):*
> ```Smalltalk
>         }
> 
>         if (projectVersifications.Count > 1)
> ```

I don't think this is correct. It is checking the number of projects, not the number of versifications.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 415 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeuCuG7AWtJaJXKY1W:-OyeuCuG7AWtJaJXKY1X:b-2i0j9g) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L415)):*
> ```Smalltalk
>                     Severity = BuildDiagnosticSeverity.Info,
>                     Message =
>                         $"There are multiple versifications represented among Paratext projects selected for training or inferencing: {string.Join(", ", projectVersifications)}.",
> ```

Are you intending to print out the dictionary entries?
___
*[`src/Serval/src/Serval.WordAlignment/Dtos/WordAlignmentExecutionDataDto.cs` line 14 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyeyh5t7J7WNG_FJsK-:-Oyeyh5t7J7WNG_FJsK0:b5s0nq9) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.WordAlignment/Dtos/WordAlignmentExecutionDataDto.cs#L14)):*
> ```Smalltalk
>     [Obsolete]
>     public IReadOnlyList<string> Warnings { get; init; } = [];
>     public IReadOnlyList<DiagnosticDto> Diagnostics { get; init; } = [];
> ```

This should be optional.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs` line 27 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyewC2M9fGC1n-bwpZI:-OyewC2M9fGC1n-bwpZJ:b2g6bok) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs#L27)):*
> ```Smalltalk
>     private readonly ILanguageTagService _languageTagService = languageTagService;
>     private const string ModelName = "NLLB";
>     private const string MinimumTrainCount = "600"; //TODO move to options?
> ```

This should be an `int` and probably be an option.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs` line 234 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyewdB-DcgLdtZQK_4G:-OyewdB-DcgLdtZQK_4H:b2pgsdq) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs#L234)):*
> ```Smalltalk
>                     Data = new Dictionary<string, object>
>                     {
>                         { "resolvedCode", sourceLanguageTag },
> ```

Is this supposed to be `resolvedCode` instead of `sourceLanguageTag`?
___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs` line 253 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyewljF5ZSbH3FSxUsR:-OyewljF5ZSbH3FSxUsS:brpu588) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs#L253)):*
> ```Smalltalk
>                     Data = new Dictionary<string, object>
>                     {
>                         { "resolvedCode", targetLanguageTag },
> ```

Is this supposed to be `resolvedCode` instead of `targetLanguageTag`?
___
*[`src/Serval/src/Serval.WordAlignment/Dtos/DiagnosticDto.cs` line 3 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyezZXc8JVwTJQ602kV:-OyezZXc8JVwTJQ602kW:b272q0n) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.WordAlignment/Dtos/DiagnosticDto.cs#L3)):*
> ```Smalltalk
> namespace Serval.WordAlignment.Dtos;
> 
> public record DiagnosticDto
> ```

This should be called `WordAlignmentDiagnosticDto`.

Enkidu93 (commented):
> @Enkidu93 made 19 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: all files reviewed, 19 unresolved discussions (waiting on ddaspit).

___
*[a discussion](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeytAJ81ziYsTyMsL_:-Oyj29qmEFCr8U_ohEGE:b-l7hxhz) (no related file):*

Are you planning on adding unit tests? And yes, we will want to add a page for the diagnostics to the wiki page. It would probably be good to have a link to the wiki page in the Swagger docs.
</blockquote></details>

Yep, I don't know if you saw the high-level comment I left when I opened the PR. Those are among the TODOs. (I know sometimes Reviewable seems to bury those top-level initial comments which is frustrating.) I wasn't sure exactly how we want to handle the wiki: I figured once the code and document have both been reviewed, I'll basically copy the doc into the wiki and link it in the `StartBuild` Swagger docs since there isn't a way, I don't think, to make the wiki update part of this PR per se. I will work on tests though - although I expect those will be a little hairy. 
___
*[`src/Machine/src/Serval.Machine.Shared/Services/EchoPreprocessBuildJob.cs` line 59 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeyFmQ2CPfpKMcs9Md:-Oyj38ieBBzfNF_Qh3_Z:bb6un1j) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/EchoPreprocessBuildJob.cs#L59)):*

Make the parameter name explicit.
</blockquote></details>

Done
___
*[`src/Machine/src/Serval.Machine.Shared/Services/EchoWordAlignmentPreprocessBuildJob.cs` line 119 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeyKqm6nIOpIgMBaOa:-Oyj3Kdr9sBARSGBAvAH:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/EchoWordAlignmentPreprocessBuildJob.cs#L119)):*

Make the parameter name explicit.
</blockquote></details>

Done.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs` line 27 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyewC2M9fGC1n-bwpZI:-Oyj3PO210kcBs538aRe:be3wxss) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs#L27)):*

This should be an `int` and probably be an option.
</blockquote></details>

OK, I made it an int. In general, how do we want to handle these options? I don't like having the model name in two classes for example. I suppose we could have some kind of diagnostics factory that has these options configured. It could have a method like `CreateDiagnostic(string code, Dictionary<string, object> data)` and we could validate all the data parameters there. Then we could also more easily test the factory with unit tests and then just confirm that the factory method is being called correctly with a mock in the build job classes. I think it also would make it easier to add new diagnostics. The only tricky part would be that this service would need to be called from both `PlatformService` (or at least that's the natural place) and the build job classes and potentially other places in the future. I'm not sure where a class like this would live. What do you think?
___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs` line 234 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyewdB-DcgLdtZQK_4G:-OyjGAL01wGnCx-vdJgQ:bphtzfw) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs#L234)):*

Is this supposed to be `resolvedCode` instead of `sourceLanguageTag`?
</blockquote></details>

Yes, good catch. Something unit tests will catch hopefully :zany_face:. 
___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs` line 253 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyewljF5ZSbH3FSxUsR:-OyjGEoc97Ezp78n_jwJ:b60r506) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs#L253)):*

Is this supposed to be `resolvedCode` instead of `targetLanguageTag`?
</blockquote></details>

Yes, done.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 203 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevFYvA9U8UOSbV3gc:-OyjGH4aE3xfKfreUkOp:b3jl6wc) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L203)):*

We have a `MaxWarnings`. We should probably have a `MaxDiagnostics`.
</blockquote></details>

OK, as long as it's a pretty high value, I think that's fine. I figure we'll want to sort by severity before limiting. The only thing that's a little weird is adding a diagnostic that means there were too many diagnostics :thinking: - I'm not sure I like that.  What do you think? Maybe it's fine.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 225 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevWuEA_DKevPhvv4g:-OyjAcoM1BYfvf1rDBz1:bx7ki4x) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L225)):*

I don't love that we are running `AnalyzeUsfmVersification` twice for a build. Could we do this once and pass it to `GetWarnings` and `GetDiagnostics` or derive the warnings from the diagnostics?
</blockquote></details>

We could. The only reason I did it this way was so that the warnings and diagnostics were completely separate. That way we could strip out the warnings easily when the time comes without affecting the diagnostics and adding the diagnostics wouldn't affect the warnings. But I could just make the warnings a list of all the diagnostic messages. 
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 243 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyev57EBtxN-xVjeh55:-OyjAyKq9D3pUQWwkaSR:bqokqid) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L243)):*

This is probably overkill, but I would feel a bit less anxious if the number part of the code was 4 digits.
</blockquote></details>

OK :laughing:, I can do that. 
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 406 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyetZcA6M1ar9SzQaQe:-OyjGV4nCh1h9LaFtCq7:b-5aozdg) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L406)):*

I don't think this is correct. It is checking the number of projects, not the number of versifications.
</blockquote></details>

Yes, sorry, I switched to a map since Nathaniel asked but didn't update this properly. Done.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 415 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeuCuG7AWtJaJXKY1W:-OyjGklRBMPKoowVPLDj:b8hwtut) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L415)):*

Are you intending to print out the dictionary entries?
</blockquote></details>

I think we probably ought to. What do you think? I can format them another way explicitly if you prefer.
___
*[`src/Serval/src/Serval.Translation/Dtos/DiagnosticDto.cs` line 3 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyezKLa8lMam6ClM92a:-OyjHJMy-AK13NoHHUla:bp0ppvv) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Dtos/DiagnosticDto.cs#L3)):*

We should call this `TranslationDiagnosticDto`, so the name doesn't collide with the word alignment diagnostic in `Serval.Client`. `ExecutionDataDto` should really be renamed too, but that would be a breaking change at this point.
</blockquote></details>

Done. Yeah, I noticed that :/.  
___
*[`src/Serval/src/Serval.Translation/Dtos/ExecutionDataDto.cs` line 14 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeyWf8-87jOysiqwBH:-OyjH_aLD4exJJtWjx_T:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Dtos/ExecutionDataDto.cs#L14)):*

This should be optional.
</blockquote></details>

Done.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 475 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyeuqgg0CaqSK6xI6Wq:-OyjHg5lE9dfdBxZcTlh:b60r506) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L475)):*

I think this is supposed to be `MODEL-003`.
</blockquote></details>

Yes, done.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 479 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyey5ke-gDOHXvCWCYp:-OyjJB_16Va3nuBdH2Ua:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L479)):*

This confidence should be formatted.
</blockquote></details>

Done.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 488 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyexNH537JvPXlVqdHF:-OyjK-sUD2qlCf3_pOcw:b-3qgy28) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L488)):*

This deferred LINQ query is getting run multiple times.
</blockquote></details>

Done. Good catch. Just `ToList()`'d it. Do we have a preference for the new spread syntax where we can just use `ToList()`? I totally understand using it for concatenation, but I feel like it's a little less readable than, e.g., `ToList()` it situations like these.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 492 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyexu7lEjeBlB6whOUb:-OyjKXtIEshu-6GaTqzR:b-9f087b) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L492)):*

Can we perform this update and the next in a single Mongo call?
</blockquote></details>

Yep, done.
___
*[`src/Serval/src/Serval.WordAlignment/Dtos/DiagnosticDto.cs` line 3 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyezZXc8JVwTJQ602kV:-OyjKZCrBencwJoa7CCY:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.WordAlignment/Dtos/DiagnosticDto.cs#L3)):*

This should be called `WordAlignmentDiagnosticDto`.
</blockquote></details>

Done.
___
*[`src/Serval/src/Serval.WordAlignment/Dtos/WordAlignmentExecutionDataDto.cs` line 14 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-Oyeyh5t7J7WNG_FJsK-:-OyjK_FM1VeyWXzB1fg8:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.WordAlignment/Dtos/WordAlignmentExecutionDataDto.cs#L14)):*

This should be optional.
</blockquote></details>

Done.

ddaspit (commented):
> @ddaspit partially reviewed 15 files and all commit messages, made 7 comments, and resolved 15 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: all files reviewed, 5 unresolved discussions (waiting on Enkidu93).

___
*[a discussion](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeytAJ81ziYsTyMsL_:-OyjpfHD6cQ_Dplk6s3r:b-c54rzp) (no related file):*

Yep, I don't know if you saw the high-level comment I left when I opened the PR. Those are among the TODOs. (I know sometimes Reviewable seems to bury those top-level initial comments which is frustrating.) I wasn't sure exactly how we want to handle the wiki: I figured once the code and document have both been reviewed, I'll basically copy the doc into the wiki and link it in the `StartBuild` Swagger docs since there isn't a way, I don't think, to make the wiki update part of this PR per se. I will work on tests though - although I expect those will be a little hairy.
</blockquote></details>

Yes, that sounds right.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs` line 27 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyewC2M9fGC1n-bwpZI:-Oyjn1vK1CiJvrR1-ay2:b2m1yv7) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs#L27)):*

OK, I made it an int. In general, how do we want to handle these options? I don't like having the model name in two classes for example. I suppose we could have some kind of diagnostics factory that has these options configured. It could have a method like `CreateDiagnostic(string code, Dictionary<string, object> data)` and we could validate all the data parameters there. Then we could also more easily test the factory with unit tests and then just confirm that the factory method is being called correctly with a mock in the build job classes. I think it also would make it easier to add new diagnostics. The only tricky part would be that this service would need to be called from both `PlatformService` (or at least that's the natural place) and the build job classes and potentially other places in the future. I'm not sure where a class like this would live. What do you think?
</blockquote></details>

When you create the diagnostic factory in Serval, make sure that it isn't coupled to the Machine engine implementation in any way.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 203 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevFYvA9U8UOSbV3gc:-Oyjo3Tx4oK514baF00V:b-yesrt1) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L203)):*

OK, as long as it's a pretty high value, I think that's fine. I figure we'll want to sort by severity before limiting. The only thing that's a little weird is adding a diagnostic that means there were too many diagnostics :thinking: - I'm not sure I like that.  What do you think? Maybe it's fine.
</blockquote></details>

I think it is fine. We need some way to communicate that we hit the max. We certainly don't want to create a huge `Build` that we store in the DB or pass from the API endpoint.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 225 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevWuEA_DKevPhvv4g:-OyjpB2Q99-k1x1ieC_d:bfhtth2) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L225)):*

We could. The only reason I did it this way was so that the warnings and diagnostics were completely separate. That way we could strip out the warnings easily when the time comes without affecting the diagnostics and adding the diagnostics wouldn't affect the warnings. But I could just make the warnings a list of all the diagnostic messages.
</blockquote></details>

If we generate the warnings from the diagnostics, it should be easy to strip out.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 415 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeuCuG7AWtJaJXKY1W:-Oyjpx5Y-kQ39MFs71fS:bi821fk) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L415)):*

I think we probably ought to. What do you think? I can format them another way explicitly if you prefer.
</blockquote></details>

I would prefer that we explicitly formatted them even if it is basically the same as `ToString` generates for a dictionary entry.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 488 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyexNH537JvPXlVqdHF:-OyjpSOIDAEpLhAqqU6p:b-j518pl) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L488)):*

Done. Good catch. Just `ToList()`'d it. Do we have a preference for the new spread syntax where we can just use `ToList()`? I totally understand using it for concatenation, but I feel like it's a little less readable than, e.g., `ToList()` it situations like these.
</blockquote></details>

I prefer `ToList`, but I also don't like fighting with the suggestions from VS Code. I'm fine with either.
___
*[`src/Serval/src/Serval.WordAlignment/Dtos/WordAlignmentDiagnosticDto.cs` line 3 at r2](https://reviewable.io/reviews/sillsdev/serval/1012#-OyjlzE090Tw_z6RF1xW:-OyjlzE090Tw_z6RF1xX:b-7d971s) ([raw file](https://github.com/sillsdev/serval/blob/28753aa5aaa90a1c5ff15e1c79fb7c8b17fffda1/src/Serval/src/Serval.WordAlignment/Dtos/WordAlignmentDiagnosticDto.cs#L3)):*
> ```Smalltalk
> namespace Serval.WordAlignment.Dtos;
> 
> public record WordAlignmentDiagnosticDto
> ```

Thinking about this a bit more. I think this should be a shared DTO in `Serval.Shared`. The corresponding model can also be shared.

Enkidu93 (commented):
> @Enkidu93 made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: 11 of 38 files reviewed, 5 unresolved discussions (waiting on ddaspit).

___
*[`src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs` line 27 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyewC2M9fGC1n-bwpZI:-Oz7JzOEEMemt9aGyFCk:bkze45e) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/NmtPreprocessBuildJob.cs#L27)):*

When you create the diagnostic factory in Serval, make sure that it isn't coupled to the Machine engine implementation in any way.
</blockquote></details>

Let me know how this looks. There's still a question of where to move a couple of the options/parameters (the confidence threshold and the model name), but otherwise, I think this is complete. There's a bit of me that thinks the model name should just be a property of the `Build`. We've made this diagnostics generic entirely because we expect to be adding new models; and if we add new models, I don't think we want the model type just living in `Options`. I could just put the confidence threshold in the options for the platform service or maybe outsource that check to another class which would make it more extensible if we change how we threshold to something more complex (?). 
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 203 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevFYvA9U8UOSbV3gc:-Oz7KtDM6U3Zxhm_2WjR:bm3yqt3) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L203)):*

I think it is fine. We need some way to communicate that we hit the max. We certainly don't want to create a huge `Build` that we store in the DB or pass from the API endpoint.
</blockquote></details>

OK, done.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 225 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevWuEA_DKevPhvv4g:-Oz7Kuaz7JUbZUOx7mGU:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L225)):*

If we generate the warnings from the diagnostics, it should be easy to strip out.
</blockquote></details>

Done.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 415 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyeuCuG7AWtJaJXKY1W:-Oz7KviN3s6bZKE7ME5u:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L415)):*

I would prefer that we explicitly formatted them even if it is basically the same as `ToString` generates for a dictionary entry.
</blockquote></details>

Done.
___
*[`src/Serval/src/Serval.Translation/Services/PlatformService.cs` line 488 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyexNH537JvPXlVqdHF:-Oz7KwkR25fMDwZtcJtj:b-egcii0) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Serval/src/Serval.Translation/Services/PlatformService.cs#L488)):*

I prefer `ToList`, but I also don't like fighting with the suggestions from VS Code. I'm fine with either.
</blockquote></details>

Yeah, me too. OK. 
___
*[`src/Serval/src/Serval.WordAlignment/Dtos/WordAlignmentDiagnosticDto.cs` line 3 at r2](https://reviewable.io/reviews/sillsdev/serval/1012#-OyjlzE090Tw_z6RF1xW:-Oz7KyxtB49KjZhN3-wE:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/28753aa5aaa90a1c5ff15e1c79fb7c8b17fffda1/src/Serval/src/Serval.WordAlignment/Dtos/WordAlignmentDiagnosticDto.cs#L3)):*

Thinking about this a bit more. I think this should be a shared DTO in `Serval.Shared`. The corresponding model can also be shared.
</blockquote></details>

Done.

ddaspit (changes requested):
> @ddaspit partially reviewed 27 files and all commit messages, made 3 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: all files reviewed, 2 unresolved discussions (waiting on Enkidu93 and pmachapman).

___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 203 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevFYvA9U8UOSbV3gc:-Oz8VG2882sMR6VnWOMR:b-zeb4x2) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L203)):*

OK, done.
</blockquote></details>

The only other option would be to add a flag to `ExecutionData` that indicates that the max was hit. It might be cleaner. @pmachapman What do you think would be best from Scripture Forge's perspective?
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 310 at r3](https://reviewable.io/reviews/sillsdev/serval/1012#-Oz8YdaZ3QxqTao1EC8b:-Oz8YdaZ3QxqTao1EC8c:bugjhwq) ([raw file](https://github.com/sillsdev/serval/blob/0a7166bb0916bbeb2ed2112375818828d7be0aae/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L310)):*
> ```Smalltalk
>         }
>         return diagnostics
>             .Select(d => new BuildDiagnostic
> ```

I don't love how we create the Serval contracts, convert them to Machine models, then convert them back to Serval contracts. The right direction would be to create the Machine models, then convert them to the Serval contracts. Or get rid of the Machine models altogether and just pass the Serval contracts.
___
*[`src/Serval/src/Serval.Shared/Services/BuildDiagnosticService.cs` line 26 at r3](https://reviewable.io/reviews/sillsdev/serval/1012#-Oz8U-GR-2gKxPw1K4W-:-Oz8U-GS8is7cM84zP3t:b32ocv4) ([raw file](https://github.com/sillsdev/serval/blob/0a7166bb0916bbeb2ed2112375818828d7be0aae/src/Serval/src/Serval.Shared/Services/BuildDiagnosticService.cs#L26)):*
> ```Smalltalk
> 
>     private static readonly Dictionary<string, DiagnosticInfo> Diagnostics = new()
>     {
> ```

I like that this provides a clear spec of the diagnostics along with a way to validate the data field.

Enkidu93 (commented):
> @Enkidu93 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: all files reviewed, 2 unresolved discussions (waiting on ddaspit and pmachapman).

___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 203 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevFYvA9U8UOSbV3gc:-Oz8bO6VEBVkM0mfvm1B:b-oqq8a4) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L203)):*

The only other option would be to add a flag to `ExecutionData` that indicates that the max was hit. It might be cleaner. @pmachapman What do you think would be best from Scripture Forge's perspective?
</blockquote></details>

Yeah, I'd prefer that. We'll see what Peter thinks.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 310 at r3](https://reviewable.io/reviews/sillsdev/serval/1012#-Oz8YdaZ3QxqTao1EC8b:-Oz8bRMkEFUddDFeC-Ax:b-goxtm) ([raw file](https://github.com/sillsdev/serval/blob/0a7166bb0916bbeb2ed2112375818828d7be0aae/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L310)):*

I don't love how we create the Serval contracts, convert them to Machine models, then convert them back to Serval contracts. The right direction would be to create the Machine models, then convert them to the Serval contracts. Or get rid of the Machine models altogether and just pass the Serval contracts.
</blockquote></details>

Me neither. The problem is that we can't have a Serval service creating  Machine models. You could pass the contract models themselves but then that'd require the `BuildExecutionData` having the contract as the type of one of its properties. My consolation in this back and forth is that we do have the contracts working both directions (e.g. the `UsfmVersificationDiagnosticContract` goes Serval -> Machine whereas `BuildProgressStatusContract` goes the other way). This contract is just going both directions: Calculated by Serval, sent to Machine, processed, and then ultimately used to communicate back to Serval. And we have to have this service in Serval.Shared because it is used in both Serval and Machine. I'm not sure what we can do. I suppose we could make to contracts to separate the two ways of communication, but they would be identical.
___
*[`src/Serval/src/Serval.Shared/Services/BuildDiagnosticService.cs` line 26 at r3](https://reviewable.io/reviews/sillsdev/serval/1012#-Oz8U-GR-2gKxPw1K4W-:-Oz8eGDP7F87PiuJYCo4:blal4ws) ([raw file](https://github.com/sillsdev/serval/blob/0a7166bb0916bbeb2ed2112375818828d7be0aae/src/Serval/src/Serval.Shared/Services/BuildDiagnosticService.cs#L26)):*

I like that this provides a clear spec of the diagnostics along with a way to validate the data field.
</blockquote></details>

Me too. I wish that it were strongly typed such that you were forced to supply the correct data types and it'd fail at compile time otherwise, but I don't see a way to do that without making a class for each diagnostic.

pmachapman (commented):
> @pmachapman made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: all files reviewed, 2 unresolved discussions (waiting on ddaspit).

___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 203 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevFYvA9U8UOSbV3gc:-Oz8sXhC0TwtVZ8TjNjf:bny5zaa) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L203)):*

Yeah, I'd prefer that. We'll see what Peter thinks.
</blockquote></details>

Either option is fine, I think.

ddaspit (commented):
> @ddaspit reviewed 21 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93).

___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 310 at r3](https://reviewable.io/reviews/sillsdev/serval/1012#-Oz8YdaZ3QxqTao1EC8b:-OzBttab0RrEeh-dQV1I:bpv63jq) ([raw file](https://github.com/sillsdev/serval/blob/0a7166bb0916bbeb2ed2112375818828d7be0aae/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L310)):*

Me neither. The problem is that we can't have a Serval service creating  Machine models. You could pass the contract models themselves but then that'd require the `BuildExecutionData` having the contract as the type of one of its properties. My consolation in this back and forth is that we do have the contracts working both directions (e.g. the `UsfmVersificationDiagnosticContract` goes Serval -> Machine whereas `BuildProgressStatusContract` goes the other way). This contract is just going both directions: Calculated by Serval, sent to Machine, processed, and then ultimately used to communicate back to Serval. And we have to have this service in Serval.Shared because it is used in both Serval and Machine. I'm not sure what we can do. I suppose we could make to contracts to separate the two ways of communication, but they would be identical.
</blockquote></details>

We should just remove the `Diagnostic` model and pass the Serval contracts directly. `BuildExecutionData.Diagnostics` can just use the contract type. That is better than converting back and forth.

Enkidu93 (commented):
> @Enkidu93 made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: 22 of 74 files reviewed, 1 unresolved discussion (waiting on ddaspit and pmachapman).

___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 203 at r1](https://reviewable.io/reviews/sillsdev/serval/1012#-OyevFYvA9U8UOSbV3gc:-OzRwnKM1I0S0Yw_F7s_:bphu25z) ([raw file](https://github.com/sillsdev/serval/blob/a3a15842977af79e97d13948a2bec2fc4d39e203/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L203)):*

Either option is fine, I think.
</blockquote></details>

I went ahead and switched to using a flag.
___
*[`src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs` line 310 at r3](https://reviewable.io/reviews/sillsdev/serval/1012#-Oz8YdaZ3QxqTao1EC8b:-OzRwsE56N1GiLd2INql:b-896fix) ([raw file](https://github.com/sillsdev/serval/blob/0a7166bb0916bbeb2ed2112375818828d7be0aae/src/Machine/src/Serval.Machine.Shared/Services/PreprocessBuildJob.cs#L310)):*

We should just remove the `Diagnostic` model and pass the Serval contracts directly. `BuildExecutionData.Diagnostics` can just use the contract type. That is better than converting back and forth.
</blockquote></details>

Done.

Enkidu93:
I've gone ahead and added tests for the diagnostics. I'm realizing in general we need to add some more testing for word alignment now that they are separate, but that's not new with this PR, so if it's OK with you, I'd prefer addressing that separately since this PR has already snowballed quite a bit.

Enkidu93:
I went ahead and added a link to the wiki. I kept it very short in the Swagger docs since we really don't have any other information regarding things in the execution data. I can elaborate if you think it's necessary but then I should probably elaborate on other things as well. (We could create an issue?). I will work on adding what's in the doc to the wiki, but I don't think that will take long or needs to hold up this PR.

ddaspit (commented):
> @ddaspit reviewed 52 files and all commit messages, made 4 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: all files reviewed, 4 unresolved discussions (waiting on Enkidu93).

___
*[`src/Machine/src/Serval.Machine.Translation/Services/ServalTranslationPlatformService.cs` line 109 at r5](https://reviewable.io/reviews/sillsdev/serval/1012#-OzROgca4vcBTloSfeTO:-OzROgca4vcBTloSfeTP:b-bk4zk6) ([raw file](https://github.com/sillsdev/serval/blob/1dfbc9aff69215312f308cf908d4cc27c885ab79/src/Machine/src/Serval.Machine.Translation/Services/ServalTranslationPlatformService.cs#L109)):*
> ```Smalltalk
>                 Warnings = executionData.Warnings,
>                 Diagnostics = executionData
>                     .Diagnostics?.Select(d => new DiagnosticContract
> ```

Unless I'm missing something, I don't think this is still necessary.
___
*[`src/Machine/src/Serval.Machine.WordAlignment/Services/ServalWordAlignmentPlatformService.cs` line 109 at r5](https://reviewable.io/reviews/sillsdev/serval/1012#-OzRR5t_7sVgMd8Ql9Te:-OzRR5t_7sVgMd8Ql9Tf:b-bk4zk6) ([raw file](https://github.com/sillsdev/serval/blob/1dfbc9aff69215312f308cf908d4cc27c885ab79/src/Machine/src/Serval.Machine.WordAlignment/Services/ServalWordAlignmentPlatformService.cs#L109)):*
> ```Smalltalk
>                 Diagnostics =
>                     executionData
>                         .Diagnostics?.Select(d => new DiagnosticContract
> ```

Unless I'm missing something, I don't think this is still necessary.
___
*[`src/Serval/src/Serval.Translation/Models/BaseModel.cs` line 14 at r5](https://reviewable.io/reviews/sillsdev/serval/1012#-OzRNkzo4b1oPYzx1KO8:-OzRNkzo4b1oPYzx1KO9:bqm83zt) ([raw file](https://github.com/sillsdev/serval/blob/1dfbc9aff69215312f308cf908d4cc27c885ab79/src/Serval/src/Serval.Translation/Models/BaseModel.cs#L14)):*
> ```Smalltalk
>         return model switch
>         {
>             BaseModel.NLLB => "facebook/nllb-200-distilled-1.3B",
> ```

The Machine engine should define this mapping.
___
*[`src/Serval/src/Serval.Translation/Models/Build.cs` line 19 at r5](https://reviewable.io/reviews/sillsdev/serval/1012#-OzRNIud3hgZljqZxq9T:-OzRNIud3hgZljqZxq9U:bbck2a4) ([raw file](https://github.com/sillsdev/serval/blob/1dfbc9aff69215312f308cf908d4cc27c885ab79/src/Serval/src/Serval.Translation/Models/Build.cs#L19)):*
> ```Smalltalk
>     public DateTime? DateFinished { get; init; }
>     public IReadOnlyDictionary<string, object>? Options { get; init; }
>     public BaseModel? BaseModel { get; init; }
> ```

This should be a string. Defining the base model values in Serval is a bit of a leaky abstraction. The Machine engine should define and validate the value. The values can be defined as string constants.

Enkidu93 (commented):
> @Enkidu93 made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012)* status: 50 of 82 files reviewed, 4 unresolved discussions (waiting on ddaspit and pmachapman).

___
*[`src/Machine/src/Serval.Machine.Translation/Services/ServalTranslationPlatformService.cs` line 109 at r5](https://reviewable.io/reviews/sillsdev/serval/1012#-OzROgca4vcBTloSfeTO:-OzT1O1P6k4HVrKxqBtC:b31eocu) ([raw file](https://github.com/sillsdev/serval/blob/1dfbc9aff69215312f308cf908d4cc27c885ab79/src/Machine/src/Serval.Machine.Translation/Services/ServalTranslationPlatformService.cs#L109)):*

Unless I'm missing something, I don't think this is still necessary.
</blockquote></details>

Yep, sorry, done.
___
*[`src/Machine/src/Serval.Machine.WordAlignment/Services/ServalWordAlignmentPlatformService.cs` line 109 at r5](https://reviewable.io/reviews/sillsdev/serval/1012#-OzRR5t_7sVgMd8Ql9Te:-OzT1PM40tSuCiVzzjg-:b60r506) ([raw file](https://github.com/sillsdev/serval/blob/1dfbc9aff69215312f308cf908d4cc27c885ab79/src/Machine/src/Serval.Machine.WordAlignment/Services/ServalWordAlignmentPlatformService.cs#L109)):*

Unless I'm missing something, I don't think this is still necessary.
</blockquote></details>

Yes, done.
___
*[`src/Serval/src/Serval.Translation/Models/Build.cs` line 19 at r5](https://reviewable.io/reviews/sillsdev/serval/1012#-OzRNIud3hgZljqZxq9T:-OzT1Rb59eIgu3h8pYwG:byzozny) ([raw file](https://github.com/sillsdev/serval/blob/1dfbc9aff69215312f308cf908d4cc27c885ab79/src/Serval/src/Serval.Translation/Models/Build.cs#L19)):*

This should be a string. Defining the base model values in Serval is a bit of a leaky abstraction. The Machine engine should define and validate the value. The values can be defined as string constants.
</blockquote></details>

That makes sense. I guess we should list the supported models in the swagger documentation and/or in the wiki? Let me know if I've carried this out as you envisioned.
___
*[`src/Serval/src/Serval.Translation/Models/BaseModel.cs` line 14 at r5](https://reviewable.io/reviews/sillsdev/serval/1012#-OzRNkzo4b1oPYzx1KO8:-OzT2ITc1QB7gmU22k-Q:bs7wumx) ([raw file](https://github.com/sillsdev/serval/blob/1dfbc9aff69215312f308cf908d4cc27c885ab79/src/Serval/src/Serval.Translation/Models/BaseModel.cs#L14)):*

The Machine engine should define this mapping.
</blockquote></details>

Done.

ddaspit (approved):
I have one comment to change the name of the `BaseModel` property to `Model`. I'm going to approve now, so you don't have to wait for me to approve after you make the change.

> @ddaspit partially reviewed 32 files and all commit messages, made 2 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1012#-:-OzT6Zfg7wOIvAWT7Vdc:btnwn2)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Enkidu93).

___
*[`src/Serval/src/Serval.Translation/Models/Build.cs` line 19 at r5](https://reviewable.io/reviews/sillsdev/serval/1012#-OzRNIud3hgZljqZxq9T:-OzT5raQElFCG8qtK_uO:btvj2s9) ([raw file](https://github.com/sillsdev/serval/blob/1dfbc9aff69215312f308cf908d4cc27c885ab79/src/Serval/src/Serval.Translation/Models/Build.cs#L19)):*

That makes sense. I guess we should list the supported models in the swagger documentation and/or in the wiki? Let me know if I've carried this out as you envisioned.
</blockquote></details>

This looks good. I would rename the property to `Model`. Depending on the engine, the model might not necessarily be a base model, e.g. LLMs that we don't fine tune.

PR #1021, opened 2026-08-06
Use correct corpus versification for chapter calculation

### LINKED ISSUE #1013: Unable to generate chapters from a source with non-canonical chapters
I was trying to generate a draft for Psalms and specified that I wanted Psalm 151 which is in the source project that I used Russian: Synodal Orthodox Version 1994 (synod). I received a 400 response from Serval explaining that the chapter range is invalid. I am guessing that Serval has guards to ensure that the appropriate chapters were indicated and available for drafting, but in this case Psalm 151 was marked as invalid when it should not be.

Image

pmachapman:
Fixes #1013

This PR fixes a crash when drafting chapters that are not in the target versification by using the source versification when relevant.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/serval/1021)

### DIFF
--- a/src/Serval/src/Serval.Translation/Services/ContractMapper.cs
+++ b/src/Serval/src/Serval.Translation/Services/ContractMapper.cs
@@ -154,11 +154,6 @@ public class ContractMapper(
             TrainingCorpus? trainingCorpus = trainingCorpora?.GetValueOrDefault(source.Id);
             PretranslateCorpus? pretranslateCorpus = pretranslateCorpora?.GetValueOrDefault(source.Id);
 
-            string? referenceFileLocation =
-                source.TargetCorpora.Count > 0 && source.TargetCorpora[0].Files.Count > 0
-                    ? Map(source.TargetCorpora[0].Files[0]).Location
-                    : null;
-
             mappedParallelCorpora.Add(
                 new ParallelCorpusContract
                 {
@@ -171,7 +166,6 @@ public class ContractMapper(
                                 sc,
                                 trainingCorpus?.SourceFilters?.Where(sf => sf.CorpusRef == sc.Id).FirstOrDefault(),
                                 pretranslateCorpus?.SourceFilters?.Where(sf => sf.CorpusRef == sc.Id).FirstOrDefault(),
-                                referenceFileLocation,
                                 trainOnAllCorpora
                                     || (trainingCorpus is not null && trainingCorpus.SourceFilters is null),
                                 pretranslateAllCorpora
@@ -187,7 +181,6 @@ public class ContractMapper(
                                 tc,
                                 trainingCorpus?.TargetFilters?.Where(sf => sf.CorpusRef == tc.Id).FirstOrDefault(),
                                 null,
-                                referenceFileLocation,
                                 trainOnAllCorpora
                                     || (trainingCorpus is not null && trainingCorpus.TargetFilters is null),
                                 pretranslateAllCorpora || pretranslateCorpus is not null
@@ -205,11 +198,11 @@ public class ContractMapper(
         MonolingualCorpus inputCorpus,
         ParallelCorpusFilter? trainingFilter,
         ParallelCorpusFilter? pretranslateFilter,
-        string? referenceFileLocation,
         bool trainOnAll,
         bool pretranslateAll
     )
     {
+        string? referenceFileLocation = inputCorpus.Files.Count > 0 ? Map(inputCorpus.Files[0]).Location : null;
         Dictionary<string, HashSet<int>>? trainOnChapters = null;
         if (
             trainingFilter is not null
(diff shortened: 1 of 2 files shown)


Enkidu93 (changes requested):
This makes sense. Thank you, Peter! Will SF use this chapter filtering correctly though in cases where the source and target versifications differ? E.g., if the source corpus has a book with chapters 1, 2, 3, but the target splits 3 into 3 and 4, will the scripture range provided be BOOK1-3 on the source and BOOK1-4 on the target? A bit of me thinks that we should actually have the scripture ranges just follow the first *source* project since the source is the anchor between training and inferencing (i.e. you need source material for both). (If I'm understanding correctly, this bug would also keep you from drafting chapters that don't exist in the target, right?) Regardless, we should specify this in the documentation. I could see it either way. One would be simpler at the monolingual corpus level, but the other would be more straightforward at the build level. What do you think? 

Related to https://github.com/sillsdev/machine/issues/421.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1021#-:-OzOH52qEHqXkUOcvik0:bqqanlm)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on pmachapman).

pmachapman (commented):
> Will SF use this chapter filtering correctly though in cases where the source and target versifications differ? E.g., if the source corpus has a book with chapters 1, 2, 3, but the target splits 3 into 3 and 4, will the scripture range provided be BOOK1-3 on the source and BOOK1-4 on the target?

SF currently uses the source USFM for drafts, so will create books 1-3 in the target.

> A bit of me thinks that we should actually have the scripture ranges just follow the first _source_ project since the source is the anchor between training and inferencing (i.e. you need source material for both).

That was one option I tried (and did work - I am happy to implement that if you want). I chose to tie it to the corpus being fed into the mapper as it seemed more coherent. I figured if the target and source were training on chapters that only exist in the source, an error would be thrown and that's OK (and so I made it so the target corpus is used for versification calculation).

> (If I'm understanding correctly, this bug would also keep you from drafting chapters that don't exist in the target, right?)

That is the bug this PR addresses?

Regardless, we should specify this in the documentation.

> Yes.

I could see it either way. One would be simpler at the monolingual corpus level, but the other would be more straightforward at the build level. What do you think?

I saw this PR as a form of stop gap until https://github.com/sillsdev/machine/issues/421 is completed. My impetus to create this PR was to bubble the error into the scope of that issue, as I think the problem could be solved far better than I have done here when we can transform versifications.

> @pmachapman made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1021#-:-OzdAARZ4njd1Qw9ZH3X:b-ascenz)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on pmachapman).

pmachapman (commented):
Rewriting as blockquotes messed up....

> Will SF use this chapter filtering correctly though in cases where the source and target versifications differ? E.g., if the source corpus has a book with chapters 1, 2, 3, but the target splits 3 into 3 and 4, will the scripture range provided be BOOK1-3 on the source and BOOK1-4 on the target?

SF currently uses the source USFM for drafts, so will create books 1-3 in the target.

> A bit of me thinks that we should actually have the scripture ranges just follow the first source project since the source is the anchor between training and inferencing (i.e. you need source material for both).

That was one option I tried (and did work - I am happy to implement that if you want). I chose to tie it to the corpus being fed into the mapper as it seemed more coherent. I figured if the target and source were training on chapters that only exist in the source, an error would be thrown and that's OK (and so I made it so the target corpus is used for versification calculation).

> (If I'm understanding correctly, this bug would also keep you from drafting chapters that don't exist in the target, right?)

That is the bug this PR addresses?

> Regardless, we should specify this in the documentation.

Yes.

> I could see it either way. One would be simpler at the monolingual corpus level, but the other would be more straightforward at the build level. What do you think?

I saw this PR as a form of stop gap until  Add capacity to transform USFM into a different versification sillsdev/machine#421 is completed. My impetus to create this PR was to bubble the error into the scope of that issue, as I think the problem could be solved far better than I have done here when we can transform versifications.

> @pmachapman made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/serval/1021#-:-OzdF6J--mPRLWRu6FAj:b-vp5byk)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on pmachapman).

Enkidu93:
Alright! If you're prepared to handle this properly in SF, I'm fine with it. It is intuitive. I think the downside is that it offloads complexity to the client. I think there is a note about using the target versification in the wiki so that should be updated. Thank you, Peter!

pmachapman:
I'm closing this PR, as after discussion, it was determined that using the target versification consistently for Scripture Ranges is the preferred option, with SF needing to convert the range from the source versification to the target versification (an update to Machine is required for this - see https://github.com/sillsdev/machine/issues/470).

